PR #4, opened 2018-05-09
Document how to transition between editions

steveklabnik:
r? @alexcrichton @killercup

### DIFF
--- a/src/editions/transitioning.md
+++ b/src/editions/transitioning.md
@@ -21,18 +21,44 @@ You can get `rustfix` from crates.io. Given that you're probably using Cargo,
 you'll want to run this:
 
 ```shell
-$ cargo +nightly install cargo-fix --git https://github.com/killercup/rustfix.git
+$ cargo install cargo-fix --git https://github.com/rust-lang-nursery/rustfix
 ```
 
-And that's it! Note that you do need a nightly Cargo to use `rustfix` at the
-moment; in fact, all of the edition stuff is nightly only for now. Eventually
-this will be stable! Same with installing from `git`, for now, it's needed.
+And that's it!
 
 ## Prepare for the next edition
 
+Before we talk about how to move to the new edition, a reminder:
+
+
+* New editions may have features that are only available in the new edition.
+  For example, `async`/`await` is available in Rust 2018, but not Rust 2015.
+* It's our intention that the migration to new editions is as smooth an experience
+  as possible. It's considered a bug if it's difficult to upgrade your crate to
+  a new edition. If you have a difficult time then a bug should be filed
+  with either rustfix or Rust itself.
+
+With that out of the way, let's get into it!
+
+### The preview period
+
+First, editions have a "preview" phase. This lets you try out the new edition
+in nightly Rust. During the preview, there's an extra step you need to take
+to opt in. At the time of writing, Rust 2018 is in its preview phase.
+
+To do that, add this feature flag to your `lib.rs` or `main.rs`:
+
+```rust
+#![feature(rust_2018_preview)]
+```
+
+This will ensure that you're enabling all of the relevant features.
+
+### Running rustfix
+
 There are some lints that can help you prepare for the next edition, but
-they're not currently turned on by default. `rustfix` has your back
-though! To turn them on and have `rustfix` fix up your code, run this:
+they're not currently turned on by default. `rustfix` has your back though!
+To turn them on and have `rustfix` fix up your code, run this:
 
 ```shell
 $ cargo +nightly fix --prepare-for 2018
--- a/src/editions/transitioning.md
+++ b/src/editions/transitioning.md
@@ -1,11 +1,74 @@
 # Transitioning your code to a new edition
 
-In general, to update your code to a new edition, the process
-will look like this:
+Transitioning between editions is built around lints. Fundamentally, the
+process works like this:
 
-* Make sure your build is warning-free.
-* Update or add the `edition` key in `Cargo.toml`
-* Fix any warnings, possibly by running `rustfix`.
+* Get your code compiling with no warnings.
+* Opt in to the new edition.
+* Fix any new warnings that may result.
 
-At this time, the `edition` key does not yet work, and `rustfix` is in a
-larval state. This section will update as our story gets better here.
+Luckily, we've been working on a tool to help assist with this process,
+`rustfix`. It can take suggestions from the compiler and automatically
+re-write your code to comply with new features and idioms.
+
+> `rustfix` is still quite young, and very much a work in development. But it works
+> for the basics! We're working hard on making it better and more robust, but
+> please bear with us for now.
+
+## Installing rustfix
+
+You can get `rustfix` from crates.io. Given that you're probably using Cargo,
+you'll want to run this:
+
+```shell
+$ cargo +nightly install cargo-fix --git https://github.com/killercup/rustfix.git
+```
+
+And that's it! Note that you do need a nightly Cargo to use `rustfix` at the
+moment; in fact, all of the edition stuff is nightly only for now. Eventually
+this will be stable! Same with installing from `git`, for now, it's needed.
+
+## Prepare for the next edition
+
+There are some lints that can help you prepare for the next edition, but
+they're not currently turned on by default. `rustfix` has your back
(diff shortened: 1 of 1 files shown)


src/editions/transitioning.md
```diff
@@ -1,11 +1,74 @@
...
+## Installing rustfix
+
+You can get `rustfix` from crates.io. Given that you're probably using Cargo,
+you'll want to run this:
+
+```shell
+$ cargo +nightly install cargo-fix --git https://github.com/killercup/rustfix.git
+```
+
+And that's it! Note that you do need a nightly Cargo to use `rustfix` at the
```
alexcrichton: Technically you don't actually need a nightly Cargo for `rustfix`, it's just a nightly rustc currently. As an implementation detail though cargo-fix purely delegates to `cargo check`, which does end up transitively and effectively requiring nightly.

In any case though the `+nightly` on the `cargo install` in theory shouldn't be necessary even today!

src/editions/transitioning.md
```diff
@@ -1,11 +1,74 @@
...
+moment; in fact, all of the edition stuff is nightly only for now. Eventually
+this will be stable! Same with installing from `git`, for now, it's needed.
+
+## Prepare for the next edition
+
+There are some lints that can help you prepare for the next edition, but
```
alexcrichton: It may be worth re-emphasizing a few points here as well:

* The 2018 edition has some features that are exclusive to it, aka not available in the 2015 edition. For example async/await.
* It's our intention that the migration to the 2018 is as smooth an experience as possible. It's a technical breaking change so it must be opt-in, but we want to make it as easy as possible for you to opt in so you can get access to the new features. (hence `cargo fix`)
* It's considered a bug if it's difficult to upgrade your crate to the 2018 edition. If you have a difficult time then a bug should be filed with either rustfix or rust-lang/rust

Manishearth: Yes, I mentioned this in gitter. We should ask people to use `#![feature(crate_in_paths)]` before they trigger rustfix

 We can also change https://github.com/rust-lang/rust/blob/master/src/librustc_resolve/lib.rs#L3354-L3355 to no longer require crate_in_paths, but they need to add it anyway so it's not that useful? I'm less fond of the idea of lints that suggest broken code (which is what it is until crate_in_paths stabilizes)

alexcrichton: After some [gitter discussion](https://gitter.im/rust-lang/WG-epoch?at=5af45ceff04ce53632c8d65d) I've added a new step to the [cheat sheet](https://paper.dropbox.com/doc/Edition-preview-cheat-sheet-HhfGRmTX4mByEOWWI2P7o) where for the *preview only* users will be required to list `#![feature(rust_2018_preview)]` in their crates they wish to migrate.

steveklabnik:
I've updated this a bit. r? anyone?

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+`rustfix`. It can take suggestions from the compiler and automatically
+re-write your code to comply with new features and idioms.
+
+> `rustfix` is still quite young, and very much a work in development. But it works
+> for the basics! We're working hard on making it better and more robust, but
+> please bear with us for now.
+
+## Installing rustfix
+
+You can get `rustfix` from crates.io. Given that you're probably using Cargo,
```
killercup: > from crates.io

well, we're actually getting it from github.com right now

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+```
+
+And that's it!
+
+## Prepare for the next edition
+
+Before we talk about how to move to the new edition, a reminder:
+
+
+* New editions may have features that are only available in the new edition.
```
killercup: > New editions may have features that are only available in the new edition.

That sounds… obvious. How about

> New editions may introduce features that are available in that form in previous editions.

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+  For example, `async`/`await` is available in Rust 2018, but not Rust 2015.
+* It's our intention that the migration to new editions is as smooth an experience
+  as possible. It's considered a bug if it's difficult to upgrade your crate to
```
killercup: > difficult

That is a super opaque und undefined metric. It's cool if you don't want to be more specific, but I'd add something like

> What do we mean by "difficult"? Migration code in macros for example: The compiler often can't directly determine how the code might be used and thus can't provide a way to upgrade in a straightforward way.

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+in nightly Rust. During the preview, there's an extra step you need to take
+to opt in. At the time of writing, Rust 2018 is in its preview phase.
+
+To do that, add this feature flag to your `lib.rs` or `main.rs`:
+
+```rust
+#![feature(rust_2018_preview)]
+```
+
+This will ensure that you're enabling all of the relevant features.
```
killercup: This is currently relevant and I'd add

> Note that during the time the preview is available, we may continue to add/enable new features with this flag!

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+they're not currently turned on by default. `rustfix` has your back though!
+To turn them on and have `rustfix` fix up your code, run this:
+
+```shell
+$ cargo +nightly fix --prepare-for 2018
+```
+
+This would turn on those lints, and fix up the project for the 2018 edition.
+If there's something that `rustfix` doesn't know how to fix automatically yet,
+the warning will be printed; you'll need to fix those manually. Do so
```
killercup: > the [usual compiler] warning will be printed

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+This will ensure that you're enabling all of the relevant features.
+
+### Running rustfix
+
+There are some lints that can help you prepare for the next edition, but
+they're not currently turned on by default. `rustfix` has your back though!
+To turn them on and have `rustfix` fix up your code, run this:
+
+```shell
+$ cargo +nightly fix --prepare-for 2018
```
killercup: Assumes people use rustup -- fine but should be mentioned

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+
+## Fix new warnings
+
+To fix up these warnings, we can use `rustfix`:
+
+```shell
+$ cargo +nightly fix
+```
+
+This will fix up all of the new warnings. Congrats! You're done!
```
killercup: > This will [try to] fix

Same drill as above regarding manual fixes (can be downplayed)

src/editions/transitioning.md
steveklabnik: This is what @alexcrichton suggested above; the opposite is also true, idk 😆

killercup: Lol, true. Okay, how about this: "New editions might change the way you write Rust -- the add syntax, language, and library features but also remove others."

src/editions/transitioning.md
alexcrichton: I personally like this sort of being opaque and undefined in the sense that if you *feel* it's difficult that's a bug we want to know about and fix. It'll definitely be a lot of work for us to bottom out "what makes it difficult" in some situations, but no one should ever get the impression that an edition upgrade is difficult ideally

src/editions/transitioning.md
```diff
@@ -1,11 +1,100 @@
...
+
+[package]
+edition = '2018'
+```
+
+That `cargo-features` line should go at the very top; and `edition` goes into
+the `[package]` section. As mentioned above, right now this is a nightly-only
+feature of Cargo, so you need to enable it for things to work.
+
+At this point, your project should compile with a regular old `cargo build`.
```
alexcrichton: s/cargo build/cargo +nightly build/

alexcrichton:
Looks great, thanks @steveklabnik!

src/editions/transitioning.md
steveklabnik: Yes, I think for now I'd like to leave this vague. As we get closer to shipping, we'll know better what those pain points are.

steveklabnik:
Thanks for the review yinz <3

PR #6, opened 2018-05-16
Minor edits

brson:
The first is just removing the double use of "create". The second is opening the `impl Trait` section by saying what it is, not where it can appear.

### DIFF
--- a/src/2018/transitioning/traits/impl-trait.md
+++ b/src/2018/transitioning/traits/impl-trait.md
@@ -1,6 +1,6 @@
 # impl Trait
 
-`impl Trait` is new syntax that currently works in function signatures.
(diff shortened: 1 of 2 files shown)


src/2018/transitioning/traits/impl-trait.md
```diff
@@ -1,6 +1,6 @@
 # impl Trait
 
-`impl Trait` is new syntax that currently works in function signatures.
+`impl Trait` is new way to specify unnamed but concrete types that implement a specific trait.
```
alexcameron89: Should this be `. . .is the new way to specify. . .`?

steveklabnik: yes please!

PR #9, opened 2018-05-18
Explain how adding keywords break compat

dwijnand:
I was a little confused why adding a keyword breaks backwards compatibility. I thought maybe forwards compatibility was meant instead, until I realised how it can break backwards compatibility.

This PR is a proposition for adding an explanation.

WDYT?


Centril:
LGTM 👍

PR #13, opened 2018-05-26
Update text in raw-identifiers.md

imp:
Cleanup wording. Should be pretty straightforward.


Centril:
Looks good 👍. Thanks!

PR #15, opened 2018-06-02
Dedup the Rust 2018 module system quote

dwijnand:
The quote "while there are simple and consistent rules.." is also
present in the section about "Path clarity" (modules/path-clarity.md)
and fits nicely there.  So I've rephrased the lead in without it.

### DIFF
--- a/src/2018/transitioning/modules/index.md
+++ b/src/2018/transitioning/modules/index.md
@@ -1,13 +1,7 @@
 # Module system
 
 The module system is one of the most confusing aspects of Rust 2015 for many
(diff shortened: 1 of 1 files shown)


dwijnand:
Yeah I can see the value in driving-home the point.

The issue is more obvious in https://rust-lang-nursery.github.io/edition-guide/print.html#module-system, which is how I consumed the guide.

I should've probably opened this as an issue first.

dwijnand:
Migrated to an issue: https://github.com/rust-lang-nursery/edition-guide/issues/18.

PR #17, opened 2018-06-04
Add some version references to shipped features

### LINKED ISSUE #16: Include version references on shipped features?
Some of the changes described in guide are already stable and available in current Rust without having to opt-in, such as default match bindings, and `?` support in main and tests.

So I was thinking: should these changes (or perhaps all changes described) mention the version as of which they're available?

dwijnand:
Fixes #16

Suggested review by @Centril. WDYT?


Centril:
LGTM 👍; Thanks!

PR #22, opened 2018-06-20
Module system revisions

aturon:
Some clarifications and consistency edits.

### DIFF
--- a/src/2018/transitioning/modules/macros.md
+++ b/src/2018/transitioning/modules/macros.md
@@ -1,9 +1,7 @@
 # Macro changes
 
-Now that `extern crate` is gone, there's a problem: to import macros, you
-need to put `#[macro_use]` on the `extern crate` declaration.
-
-To fix this issue, you can use `use` to import macros from extern crates.
+In Rust 2018, you can use `use` to import specific macros from extern crates,
+rather than the old `#[macro_use]` attribute.
 
 For example, consider a `bar` crate that implements a `baz!` macro. In
 `src/lib.rs`:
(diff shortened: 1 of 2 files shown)


src/2018/transitioning/modules/macros.md
```diff
@@ -1,9 +1,7 @@
 # Macro changes
 
-Now that `extern crate` is gone, there's a problem: to import macros, you
-need to put `#[macro_use]` on the `extern crate` declaration.
-
-To fix this issue, you can use `use` to import macros from extern crates.
+In Rust 2018, you can use `use` to import specific macros from extern crates,
```
Centril: The word "use" is repeated twice here, which feels a bit sub-optimal. Not sure what to replace it with tho.

Centril: Replace `extern crates` with `external crates` :)

otavio: What about:

In Rust 2018, the `use` keyword allows the importing of specific macros from external crates,

Centril: If we reorder:
> In Rust 2018, specific macros from external crates can be imported with `use`.

src/2018/transitioning/modules/macros.md
```diff
@@ -43,4 +43,4 @@ This moves `macro_rules` macros to be a bit closer to other kinds of items.
 ## More details
 
 This only works for `extern crate`; if you write a macro in the same crate,
-nothing changes.
\ No newline at end of file
+nothing changes from Rust 2015.
```
Mark-Simulacrum: I thought use_extern_macros permits using macros from within a crate as well. I seem to recall recent discussions toward this being a change we're making, but I'm not completely up to date...

aturon:
Review addressed.

PR #24, opened 2018-06-21
Error handling edits

aturon:
Just a few minor consistency tweaks.

### DIFF
--- a/src/2018/transitioning/errors/question-mark.md
+++ b/src/2018/transitioning/errors/question-mark.md
@@ -1,6 +1,6 @@
 # `?` in `fn main()` and `#[test]`s
 
-Typically in Rust, error handling revolves around returning `Result` and using
+Rust's error handling revolves around returning `Result` and using
 `?` to propagate errors. For those who write many small programs and, hopefully,
 many tests, one common paper cut has been mixing entry points such as `main`
 and `#[test]`s with error handling.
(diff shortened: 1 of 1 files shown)


src/2018/transitioning/errors/question-mark.md
```diff
@@ -1,6 +1,6 @@
 # `?` in `fn main()` and `#[test]`s
 
-Typically in Rust, error handling revolves around returning `Result` and using
+Rust's error handling revolves around returning `Result` and using
```
Centril: Is this not overstating things? `Option<T>` can work well in some cases as well.
Perhaps the simplification is useful for the purposes of teaching tho.

src/2018/transitioning/errors/question-mark.md
```diff
@@ -51,15 +53,15 @@ fn main() {
 ```
 
 However, in this case, the `run` function has all the interesting logic and
-`main` is just boilerplate. When you write many `#[test]`s, this can become
-a problem.
-
-As of [1.26](main-can-return-a-result), you can instead let your `#[test]`s and `main` functions return
```
Centril: Hmm... this gets rid of the link to 1.26... should we not keep it?

src/2018/transitioning/errors/question-mark.md
steveklabnik: Option isn't about error handling, IMHO.

PR #33, opened 2018-06-23
impl-trait: consistently use old then new style

### LINKED ISSUE #32: Standardize on old way then new way or vice versa
For example, in `2018/transitioning/traits/impl-trait.md`, it switches between first `impl Trait` then `T: Trait` to the other way around. I didn't read the sentence right after the code block, so I was confused about which is former and which is latter.

It would probably be good to decide on one way or the other. I think old way then new way is a better general rule than the other way around. (of course being free to change the order if it reads better.)

Hello71:
Fixes #32.


Centril:
Thank you!

PR #39, opened 2018-06-23
Use dyn Trait over the old syntax

Mark-Simulacrum:
@est31 pointed this out on IRC


Centril:
cc @steveklabnik 

This might be intentional?
The `impl Trait` section is before `dyn Trait` and has therefore not introduced the `dyn Trait` syntax yet.
Might be good avoiding introducing two things at once.

est31:
Maybe the section order can be flipped?

Centril:
@est31 I'll wait a bit for Steve's input first, but that seems reasonable :)

steveklabnik:
That was my intention, but however, it's probably best to do this, but also add a little

    > Note: see the `dyn Trait` section for an explanation of that syntax

maybe?

Centril:
@steveklabnik SGTM 👍

Mark-Simulacrum:
Okay, I've flipped the order here.

Centril:
LGTM, thanks!

PR #43, opened 2018-06-24
Is `'arg'` missing in Rust 2018 example?

jsgf:
It looks to me like Rust 2018 example is missing `'arg` in `arg: &str`.

OTOH, if the code is correct, then I'm misunderstanding how this is supposed to work which probably means it needs some clarification.

### DIFF
--- a/src/2018/transitioning/ownership-and-lifetimes/lifetime-elision-in-impl.md
+++ b/src/2018/transitioning/ownership-and-lifetimes/lifetime-elision-in-impl.md
@@ -47,6 +47,6 @@ impl MyStruct<'a> {
     fn foo(&self) -> &'a str
 
     // we can refer to 'arg, rather than conflicting with 'a
-    fn bar(&self, arg: &str) -> &'arg str
+    fn bar(&self, arg: &'arg str) -> &'arg str
 }
 ```


Mark-Simulacrum:
I believe the lifetime comes from the name of the argument; cc @aturon

steveklabnik:
That's my understanding as well.

jsgf:
That's a little confusing, since parameters don't technically have names - it's a simple pattern binding. Does that mean if you had a more complex parameter, say `&MyStruct { foo, bar }: &MyStruct` you would need to explicitly declare a lifetime? (Hm, does match ergonomics come into play here?)

(Is this another instance of "initialize struct fields from locals with the same name" - ie, mixes two logically distinct things in a way that's convenient most of the time?)

My reading of [in-band lifetimes](https://rust-lang-nursery.github.io/edition-guide/2018/transitioning/ownership-and-lifetimes/in-band-lifetimes.html), which I thought was in play here, is that you can start referencing lifetimes in the parameter list without declaring them, but there's no implication that the lifetime name comes from the parameter name.

aturon:
Fixed in an earlier PR. Thanks!

PR #51, opened 2018-06-26
add Minimum Edition to Feature Status items in Lang table

Centril:
+ Housekeeping
+ Adds the minimum edition to each lang tables in the feature status page.

### DIFF
--- a/src/2018/status.md
+++ b/src/2018/status.md
@@ -2,44 +2,81 @@
 
 ## Language
 
-| **Feature** | **Status** |
-| ----------- | ---------- |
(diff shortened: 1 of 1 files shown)


Centril:
cc @steveklabnik, @aturon -- is this information we want to advertise?

moxian:
Since I was one of the people heavily pushing for this: The motivation for the change is that the difference between "features that have landed in rust since 2015" and "features that need an `editon = "2018"` opt in" is very muddled.
It's unclear whether the features which work in current stable rust (say, `impl Trait`) will continue to work once the edition choice lands - those are advertised as changes for `edition = "2018"`, and have a "need to explicitly opt in" kind of disclaimer.

By listing the minimal edition required the confusion is gone. It's now clear - that, yes, `impl Trait` will continue to work in rust stable just fine.

cc @Lokathor

aturon:
Thanks, this is great!

PR #52, opened 2018-06-26
Small clarification to edition compatibility

eminence:
The old text ("The Rust compiler can link crates of any editions together") made it sound like old compilers would somehow magically be able to support language features from newer editions.  

The goal of this new text is to make it more clear that a given compiler release will only be able to work with the specific editions that it supports.

### DIFF
--- a/src/editions/index.md
+++ b/src/editions/index.md
@@ -27,9 +27,11 @@ This serves different purposes for different people:
 When a new edition becomes available in the compiler, crates must explicitly opt
 in to it to take full advantage. This opt in enables editions to contain
 incompatible changes, like adding a new keyword that might conflict with
-identifiers in code, or turning warnings into errors. The Rust compiler can link
-crates of any editions together. Edition changes only affect the way the
-compiler initially parses the code. Therefore, if you're using Rust 2015, and
+identifiers in code, or turning warnings into errors. A Rust compiler will
+support all editions that existed prior to the compiler's release, and can link
+crates of any supported editions together.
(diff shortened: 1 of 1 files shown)


Centril:
cc @Mark-Simulacrum ... does this sufficiently solve #53? IMO it does.

src/editions/index.md
```diff
@@ -27,9 +27,10 @@ This serves different purposes for different people:
 When a new edition becomes available in the compiler, crates must explicitly opt
 in to it to take full advantage. This opt in enables editions to contain
 incompatible changes, like adding a new keyword that might conflict with
-identifiers in code, or turning warnings into errors. The Rust compiler can link
-crates of any editions together. Edition changes only affect the way the
-compiler initially parses the code. Therefore, if you're using Rust 2015, and
+identifiers in code, or turning warnings into errors. A Rust compiler can support
+multiple editions, and can link crates of any supported editions together.
```
Centril: It would be good to say explicitly something akin to "A Rust compiler will support all editions that existed prior to the compiler's release" as this makes it sound like a Rust compiler can pick and chose.

eminence:
Thanks for the ping!  I didn't see Centril's feedback.  I pushed a new version, and rebased on upstream.

Old text:

> The Rust compiler can link crates of any editions together.

New text:

> A Rust compiler will support all editions that existed prior to the compiler's release, and can link crates of any supported editions together.

Centril:
Sounds good to me; thanks!

PR #57, opened 2018-07-09
Update lifetime-elision-in-impl.md

remexre:
An extra lifetime parameter was present; the Rust 2015 version has `SomeTrait` only taking one lifetime parameter.


Centril:
Thanks!

PR #60, opened 2018-07-18
Remove double negative and repeated word

shelbyd:
I had to reread the original sentence multiple times to understand its meaning. I think this is clearer.


Centril:
Much better, yes. Thanks!

PR #65, opened 2018-07-25
Added note about 2015 features in the feature status table

eminence:
This change falls out of some of the discussion in #63.  I don't feel particularly strongly about this change, but I do think it helps provide some useful context for this table.


Centril:
Looks good; Thanks!

PR #68, opened 2018-07-27
Document how to transition imports of procedural macros

### LINKED ISSUE #66: Document how to transition imports of procedural macros
It is not clear how code importing procedural macros using `extern crate` should be updated. A typical example:

```rust
#[macro_use] extern crate serde_derive;

#[derive(Serialize, Deserialize)]
...
```

How should `#[macro_use] extern crate` be replaced so that `#[derive(...)]` still works?

vks:
Fixes #66.

### DIFF
--- a/src/2018/transitioning/modules/macros.md
+++ b/src/2018/transitioning/modules/macros.md
@@ -30,7 +30,7 @@ Now, you write:
 
 ```rust,ignore
 // Rust 2018
-#![feature(rust_2018_preview, use_extern_macros)]
+#![feature(rust_2018_preview)]
 
 use bar::baz;
 
@@ -59,7 +59,7 @@ Now, you write instead:
 
 ```rust,ignore
 // Rust 2018
-#![feature(rust_2018_preview, use_extern_macros)]
+#![feature(rust_2018_preview)]
 use serde_derive::{Serialize, Deserialize};
 
 #[derive(Serialize, Deserialize)]
--- a/src/2018/transitioning/modules/macros.md
+++ b/src/2018/transitioning/modules/macros.md
@@ -41,6 +41,34 @@ fn main() {
 
 This moves `macro_rules` macros to be a bit closer to other kinds of items.
 
+
+## Procedural macros
+
+When using procedural macros to derive traits, you could have written:
+
+```rust,ignore
+// Rust 2015
+extern crate serde;
+#[macro_use] extern crate serde_derive;
+
+#[derive(Serialize, Deserialize)]
+struct Bar;
+```
+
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
cc @alexcrichton

src/2018/transitioning/modules/macros.md
```diff
@@ -41,6 +41,34 @@ fn main() {
...
+
+#[derive(Serialize, Deserialize)]
+struct Bar;
+```
+
+Now, you write instead:
+
+```rust,ignore
+// Rust 2018
+#![feature(rust_2018_preview, use_extern_macros)]
```
alexcrichton: I think `use_extern_macros` is implied by `rust_2018_preview`, so that feature should be removable here

src/2018/transitioning/modules/macros.md
```diff
@@ -41,6 +41,34 @@ fn main() {
...
+use serde_derive::{Serialize, Deserialize};
+
+#[derive(Serialize, Deserialize)]
+struct Bar;
+```
+
+Note that the macros you have to import have the same name as the traits you want to derive.
```
alexcrichton: I don't think this is quite true, but rather the macro's name is defined by `#[proc_macro_derive(Foo)]` and then it's simply named `Foo`. It's good practice to match a trait but it's not required!

eminence: Maybe we could add some text like:

> When importing a macro to use with custom derive, you will have to name the macro that provides the custom derive.  This generally matches the name of the trait, but check with the documentation of the crate providing the derives to be sure.  
>
> For example, here is how you would do it with serde **_insert example here_**

This has a subtle benefit of encouraging library authors to now document this :)

vks:
Unfortunately, the macros are completely missing from the documentation in the case of `serde_derive`. This seems like a rustdoc bug?

eminence:
Perhaps, but the main documentation page for serde_derive seems to be https://serde.rs/derive.html, which is manually written.  So I'm sure that the serde team will be re-writing this for Rust 2018.  

Should a new issue be opened on the rustdoc repo about how custom derive macros should be documented?

src/2018/transitioning/modules/macros.md
alexcrichton: Sounds good!

src/2018/transitioning/modules/macros.md
vks: I copied it from the existing docs, so I guess it is redundant there as well. I will remove both.

alexcrichton:
👍

PR #69, opened 2018-07-28
Explicitly mention not passing prepare flag after switching to new edition

johnthagen:
Related to https://github.com/rust-lang/cargo/issues/5778.

Based on feedback from the [internals thread](https://internals.rust-lang.org/t/help-test-out-the-2018-module-system-changes/8047/20?u=johnthagen), hopefully this clarification will help people realize not to pass the `--prepare-for` flag after they've already transitioned to the new edition.

Hopefully removes a paper cut when others test out this feature.

### DIFF
--- a/src/editions/transitioning.md
+++ b/src/editions/transitioning.md
@@ -91,10 +91,6 @@ build`. If it does not, this is a bug! Please [file an issue][issue].
 
 [issue]: https://github.com/rust-lang/rust/issues/new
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I think this might be most appropriate in the section below though? After the edition is turned on we don't necessarily recommend you keep running `cargo fix`

johnthagen:
@alexcrichton How does it look now?

alexcrichton:
💯

Issue #19, opened 2018-06-07
Suggest/Talk about --all-targets and feature gating


Mark-Simulacrum:
See https://github.com/rust-lang/rust/issues/51208#issuecomment-395573875; copying:

So `--all-targets` seems to work great. Feature handling is harder, but since rustfix supports incremental migration that's fine, for crates with non-mutually exclusive features, `cargo +nightly fix --prepare-for 2018 -- --all-targets --no-default-features` and `cargo +nightly fix --prepare-for 2018 --allow-dirty -- --all-targets --all-features` should be sufficient to migrate.

If exclusive features are present users need to run with `--feature blah` and then `--features baz`. For the alpha release that seems sufficient; we might want rustfix to automatically do this with some flag, but I don't think that's very feasible due to exclusionary feature groups.

We should update the migration docs.

Mark-Simulacrum:
@nikomatsakis [suggested](https://github.com/rust-lang/rust/issues/51208#issuecomment-395609671) that we also ask users to provide feedback about the feature patterns beyond all/none that they use.

alexcrichton:
One question that I think would be worth considering is whether to enable this sort of behavior by default as well, as 

```
cargo +nightly fix --prepare-for 2018 --allow-dirty -- --all-targets --all-features 
```

is pretty wordy!

Mark-Simulacrum:
`--all-features` is also not technically enough, as you might have features which enable/disable functionality when enabled/disabled, e.g. something like a `std` feature which is on by default but when disabled you'll see alloc or core imports, for example

Centril:
Closing since #70 rendered this moot.

PR #70, opened 2018-07-31
Migrate to using `cargo fix --edition-idioms`

alexcrichton:
This'll be a new flag being added to Cargo to help transitioning to the idioms
of the 2018 edition.

At the same time this removes the `+nightly` invocations from Cargo commands,
removes `--all-targets` as it's the default, and finally removes
`--all-features` as the default of `--all-targets` should be enough.

### DIFF
--- a/src/editions/transitioning.md
+++ b/src/editions/transitioning.md
@@ -52,7 +52,7 @@ new 2018 edition. This is where the handy `cargo fix` tool comes into the
 picture. To enable the compatibility lints for your project you run:
 
 ```shell
-$ cargo +nightly fix --prepare-for 2018 --all-targets --all-features
+$ cargo fix --edition
 ```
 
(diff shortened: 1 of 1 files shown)


Centril:
**DON'T MERGE YET!**

This PR is awaiting the stuff mentioned above to be merged into Cargo.
@alexcrichton will ping back on this PR once that is done and this is ready to be merged.

alexcrichton:
Ah yes, sorry! This is blocked on https://github.com/rust-lang/cargo/pull/5843 making its way into nightlies.

Centril:
Note to self: Once this is merged, #19 should be rendered moot so we can close that then.

alexcrichton:
Everything is now landed in rust-lang/rust and a nightly is one the way out, so merging!

PR #92, opened 2018-08-15
Use anonymous lifetime also for fmt::Formatter

niklasf:


stjepang:
@steveklabnik The guide says:

> Therefore, starting with Rust 2018, it is deprecated to leave off the lifetime parameters for non-reference-types

Is the lint not working? Because rustc is [not complaining](https://play.rust-lang.org/?version=beta&mode=debug&edition=2018&gist=ca0231b1b5bacd5e28e54c800f11184a) about elided lifetimes in types. Is this a bug or the lint was turned off for Rust 2018?

steveklabnik:
Maybe it was turned off. @Centril ?

Centril:
@steveklabnik Idiom lints are not turned on this early;

```rust
#![warn(rust_2018_idioms)]

use std::fmt;

struct StrWrap<'a>(&'a str);

fn make_wrapper<'a>(string: &'a str) -> StrWrap {
    StrWrap(string)
}

impl fmt::Debug for StrWrap<'_> {
    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
        fmt.write_str(self.0)
    }
}

fn main() {
}
```

will trigger `warning: hidden lifetime parameters in types are deprecated`.

steveklabnik:
Ah, it's an idiom lint, okay.

stjepang:
@Centril Thanks! Is the `rust_2018_idioms` lint going to be enabled in the final 1.31 release?

Centril:
@stjepang It's already provided in the 1.31 beta but as `allow`; I think we'll incrementally want to move these to `warn` by default with ~6 months for those we feel most uncertain about. But lints such as omitting `dyn`, which we feel more certain about, should probably become warn-by-default before that but not as early as 1.31 iirc.

We haven't figured out all of this yet... cc @alexcrichton for more details.

stjepang:
Thanks for clarification! I was asking because the guide changes `Formatter` to `Formatter<'_>`, which gives the impression that this is what we should write in Rust 2018, but the compiler doesn't complain about the missing lifetime. It's therefore a little confusing change.

alexcrichton:
That's the details I know of for now in that the idiom lints are all turned off by default and we'll consider turning them on over time as we become more and more confident in them

PR #93, opened 2018-08-16
Add the feature to every example & cargo +nightly

sunjay:
When I was trying the new edition, I found that I needed to add the Rust feature declaration to every example in addition to the lib.rs file and main.rs file. I also need to use `cargo +nightly fix` in order to run the commands.

This PR adds that to the edition guide.


Centril:
Thanks!

PR #94, opened 2018-08-16
fix broken link from Path Clarity to Macro Changes

durka:


Centril:
Thanks! The lack of absolute paths makes me unhappy :(

durka:
Insert joke about uniform paths here.

PR #105, opened 2018-09-17
Document deprecation of anon params

mark-i-m:
cc https://github.com/rust-lang/rust/issues/41686
cc @Centril


src/rust-2018/trait-system/no-anon-params.md
```diff
@@ -0,0 +1,20 @@
...
+parameters in trait method declarations are no longer allowed to be anonymous.
+
+For example, in the 2015 edition, this was allowed:
+```rust
+trait Foo {
+    fn foo(&self, u8);
+}
+```
+
+In the 2018 edition, all parameters must be given a variable (even if it's just
```
Centril: a variable -> an argument name;

everything else looks great :)

Centril:
Thanks!

PR #117, opened 2018-10-23
Add notes on migrating crates to work with both editions

derekdreery:
Some notes on supporting both editions while using implementation macros.

I haven't checked that the code compiles, so I would be grateful for a thorough review!

### DIFF
--- a/src/rust-2018/macros/macro-changes.md
+++ b/src/rust-2018/macros/macro-changes.md
@@ -2,6 +2,8 @@
 
 ![Minimum Rust version: beta](https://img.shields.io/badge/Minimum%20Rust%20Version-beta-orange.svg)
 
+## `macro_rules!` style macros
+
 In Rust 2018, you can import specific macros from external crates via `use`
 statements, rather than the old `#[macro_use]` attribute.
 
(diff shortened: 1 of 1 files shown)


src/rust-2018/macros/macro-changes.md
```diff
@@ -78,3 +80,158 @@ struct Bar;
...
+
+```rust,ignore
+pub struct LogLevel {
+    Warn,
+    Error
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! log {
```
dtolnay: I would rename this macro to `private_log!` or `log_impl!` or `__log!` to make it clear that importing it is not a reasonable fix (because the real log crate does export a public `log!` macro).

src/rust-2018/macros/macro-changes.md
```diff
@@ -78,3 +80,158 @@ struct Bar;
...
+So the code knows to look for any macros used locally. But wait - this won't compile, because we
+use the `stringify!` macro that isn't in our local crate (hence the convoluted example). The
+solution is to add a level of indirection: we crate a macro that wraps stringify, but is local to
+our crate. That way everything works in both editions (sadly we have to pollute the global
+namespace a bit, but that's ok).
+
+```rust,ignore
+#[doc(hidden)]
+#[macro_export]
+macro_rules! my_special_stringify {
```
dtolnay: I don't want to see `my_special_...` in an actual crate. Could you come up with a naming convention that we can recommend to all uses of this pattern?

src/rust-2018/macros/macro-changes.md
```diff
@@ -78,3 +80,158 @@ struct Bar;
...
+    ($($inner:tt)*) => {
+        stringify!($($inner)*)
```
dtolnay: Since the approach described here is intended to be picked up by all sorts of macros, I would write this call with curly braces.

```rust
    stringify! { $($inner)* }
```

Some macros cannot be forwarded with parentheses: [*playground*](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=f763e26d28e9ed7b3999f8dc7031360d). All macros can be forwarded with curly braces.

src/rust-2018/macros/macro-changes.md
```diff
@@ -78,3 +80,158 @@ struct Bar;
...
+macro_rules! log {
+    ($level:expr, $msg:expr) => {
+        println!("{}: {}", $level, $msg)
+    }
+}
+
+#[macro_export]
+macro_rules! warn {
+    ($msg:expr) => {
+        log!(stringify!($crate::LogLevel::Warn), $msg)
```
dtolnay: Please pick a different standard library macro to build this demo code around. You would never see `stringify!` called like this on a crate local path. `format_args!` could work better.

derekdreery:
I've made these changes, and turned on CI for the last example, since it should compile in all situations :)

derekdreery:
This is ready for review again.

src/rust-2018/macros/macro-changes.md
```diff
@@ -78,3 +80,192 @@ struct Bar;
...
+macro_rules! warn {
+    ($($args:tt)*) => {
+        __impl_log!($crate::LogLevel::Warn, format_args!($($args)*))
+    }
+}
+```
+
+So the code knows to look for any macros used locally. But wait - this won't compile, because we
+use the `format_args!` macro that isn't in our local crate (hence the convoluted example). The
+solution is to add a level of indirection: we crate a macro that wraps `format_args`, but is local
```
softprops: typo here `we crate` => `we create`

src/rust-2018/macros/macro-changes.md
```diff
@@ -78,3 +80,192 @@ struct Bar;
 
 This only works for macros defined in external crates.
 For macros defined locally, `#[macro_use] mod foo;` is still required, as it was in Rust 2015.
+
+### Local helper macros
```
softprops: It may be useful for authors of crates using _foreign_ helper macros to also have some examples. Those examples may almost look exactly the local examples below so it may be even better to rename this sections `helper macros` with examples using a mix of both `foreign-crate::` and `$crate::` alike.

derekdreery: Would it be possible to merge this PR first, and then address this in a new PR?

derekdreery:
Ping - I think this will be useful to rust users so we should either merge or tell me what I need to change.

Michael-F-Bryan:
I just ran into this problem when using the new `some_crate::some_macro!()` syntax and this PR would have saved me a fair amount of googling. @derekdreery is there anything left to do on this?

petrochenkov:
I'd merge this long time ago, but I don't have rights in this repo unfortunately.

ping @steveklabnik

src/rust-2018/macros/macro-changes.md
```diff
@@ -78,3 +80,192 @@ struct Bar;
...
+    ($level:expr, $msg:expr) => {{
+        println!("{}: {}", $level, $msg)
+    }}
+}
+
+/// Warn level log message
+#[macro_export(local_inner_macros)]
+macro_rules! warn {
+    ($($args:tt)*) => {
+        __impl_log!($crate::LogLevel::Warn, format_args!($($args)*))
```
Nemo157: Should this (and `error` below) be calling the internal `_log__format_args` macro?

derekdreery: Yep good spot! Thanks

derekdreery: fixed.

derekdreery:
woot

PR #125, opened 2018-12-06
aborting-on-panic.md: Typo in example config

rawler:
When trying to use the provided config, cargo warns:

```
warning: unused manifest key: profile.debug
warning: use `[profile.dev]` to configure debug builds
```


steveklabnik:
Whoops, thank you!

PR #130, opened 2018-12-11
Add 2018-specific changes.

### LINKED ISSUE #91: Include a page about 2018-only changes
Include a dedicated page that explains exactly what parts that are only available on 2018 + all the breakage.
This should be a tiny list.

ehuss:
Closes #91.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -15,6 +15,7 @@
 ## Rust 2018
 
 - [Rust 2018](rust-2018/index.md)
+    - [2018-Specific Changes](rust-2018/edition-changes.md)
     - [Module system](rust-2018/module-system/index.md)
       - [Raw identifiers](rust-2018/module-system/raw-identifiers.md)
       - [Path clarity](rust-2018/module-system/path-clarity.md)
@@ -84,4 +85,3 @@
       - [MSVC toolchain support](rust-2018/platform-and-target-support/msvc-toolchain-support.md)
       - [MUSL support for fully static binaries](rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.md)
       - [`cdylib` crates for C interoperability](rust-2018/platform-and-target-support/cdylib-crates-for-c-interoperability.md)
-    - [2018-Specific Changes](rust-2018/edition-changes.md)
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -84,3 +84,4 @@
       - [MSVC toolchain support](rust-2018/platform-and-target-support/msvc-toolchain-support.md)
       - [MUSL support for fully static binaries](rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.md)
       - [`cdylib` crates for C interoperability](rust-2018/platform-and-target-support/cdylib-crates-for-c-interoperability.md)
+    - [2018-Specific Changes](rust-2018/edition-changes.md)
--- a/src/rust-2018/edition-changes.md
+++ b/src/rust-2018/edition-changes.md
@@ -4,7 +4,7 @@ The following is a summary of changes that only apply to code compiled with
 the 2018 edition compared to the 2015 edition.
 
 ## Rust
-- [Non-lexical lifetimes].
+- [Non-lexical lifetimes]&nbsp;(future inclusion planned for 2015 edition)
 - [At most once] `?` macro repetition operator.
 - [Path changes]:
     - `use` declarations must begin with:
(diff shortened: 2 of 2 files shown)


ehuss:
Some notes:
- I have no idea where to place the page. I just stuck it at the end. Another option would be to place it directly in `rust-2018/index.md` but that didn't quite feel right.
- This is intentionally terse since it is intended to just be an overview. I'd be happy to add more detail, examples, links, etc. if you want.
- I'd be fine with dropping the Cargo section. The changes are relatively minor, and Cargo gives plenty of warnings in both editions.
- Feel free to suggest anything. This is the information *I* want to see, but I understand that the style/voice differs from the rest of the guide.
- I'll try to keep this list updated in the future. Some of these are actively being moved to 2015, and things like async/await will likely need to be added.

src/rust-2018/edition-changes.md
```diff
@@ -0,0 +1,47 @@
+# 2018-Specific Changes
+
+The following is a summary of changes that only apply to code compiled with
+the 2018 edition compared to the 2015 edition.
+
+## Rust
+- [Non-lexical lifetimes].
```
Centril: Might want to note that this is a temporary arrangement; we intend to backport later.

src/rust-2018/edition-changes.md
```diff
@@ -0,0 +1,47 @@
...
+- [At most once] `?` macro repetition operator.
+- [Path changes]:
+    - `use` declarations must begin with:
+        - `crate` – refers to the current crate.
+        - `self` – refers to the current module.
+        - `super` – refers to the parent module.
+        - An external crate name.
+        - `::` – must be followed by an external crate name. This is required
+          if an external crate has the same name as an in-scope item, to catch
+          possible ambiguities with future module changes.
```
Centril: We'll introduce uniform_paths eventually; might want to hint at that as well.

src/SUMMARY.md
```diff
@@ -83,4 +83,5 @@
       - [Global allocators](rust-2018/platform-and-target-support/global-allocators.md)
       - [MSVC toolchain support](rust-2018/platform-and-target-support/msvc-toolchain-support.md)
       - [MUSL support for fully static binaries](rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.md)
-      - [`cdylib` crates for C interoperability](rust-2018/platform-and-target-support/cdylib-crates-for-c-interoperability.md)
\ No newline at end of file
+      - [`cdylib` crates for C interoperability](rust-2018/platform-and-target-support/cdylib-crates-for-c-interoperability.md)
+    - [2018-Specific Changes](rust-2018/edition-changes.md)
```
Centril: Might want to move this to be the first item, i.e. 3.1 (and modules become 3.2).

src/rust-2018/edition-changes.md
```diff
@@ -0,0 +1,47 @@
...
+    - Paths in `pub(in path)` visibility modifiers must start with `crate`,
+      `self`, or `super`.
+- [Anonymous trait function parameters] are not allowed.
+    - [Trait function parameters] may use any irrefutable pattern when the
+      function has a body.
+- [`dyn`] is a [strict keyword], in 2015 it is a [weak keyword].
+- `async`, `await`, and `try` are [reserved keywords].
```
petrochenkov: `await` is not reserved

Centril:
```rust
error[E0721]: `await` is a keyword in the 2018 edition
 --> src/lib.rs:1:4
  |
1 | fn await() {}
  |    ^^^^^ help: you can use a raw identifier to stay compatible: `r#await`
```

You have to use `#![feature(await_macro)]` to make it unreserved.

petrochenkov: Oh god.
They future-proofed it with a hack in a lint pass.

Ok, this makes it kinda reserved in practice.

ehuss:
Good points, thanks for taking a look!

dwijnand:
This would be lovely to land, it took me quite some digging before I realised that NLL is edition 2018 only.

As a side note, I couldn't figure out how to specify the edition to `rustc` because it's not in `rustc --help` or in the guide anywhere, so I ended up looking it up in the cargo sources.  (I later found it in `rustc --help -v`.)

steveklabnik:
Thank you! People have been asking about this for a while. I think this is a fine place to put it, we can iterate if it turns out to be not ideal in practice.

dwijnand:
Now that Rust nightly is fixed, could someone rerun the merge build for this so it deploys, please?  Thank you!

https://travis-ci.com/rust-lang-nursery/edition-guide/builds/96925518

ehuss:
I poked it, but it looks like it will be queued for a long time due to other jobs running.

dwijnand:
https://rust-lang-nursery.github.io/edition-guide/rust-2018/edition-changes.html :tada:

PR #132, opened 2018-12-16
Fix link to path clarity changes section

synaestheory:


synaestheory:
Closed for #133 , which fixes more broken links.

Issue #126, opened 2018-12-07
Link broken


gurry:
The link "improved error messages" on [this page](https://rust-lang-nursery.github.io/edition-guide/rust-2018/the-compiler/index.html) is broken. The problem is due the underlying URL being this:

`https://rust-lang-nursery.github.io/edition-guide/rust-2018/the-compiler/rust-2018/the-compiler/improved-error-messages.html`

instead of this:
`https://rust-lang-nursery.github.io/edition-guide/rust-2018/the-compiler/index.html`

xonatius:
Same for other pages, for example [Module system](https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/index.html). Looks like links from 419edb885ec1a98c0747b3907003d79e3e6b93a9  need to be fixed.

ehuss:
Links should now be fixed as of #152.

PR #133, opened 2018-12-16
Links fixes

synaestheory:
Updating links to point to the docroot of the html book and including notes in README.md about using a webserver to view the output.

This addresses #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,31 +29,31 @@ $ cargo install mdbook
 
 ### Building
 
-To build the book, do this:
-
+The most straight-forward way to build and view the book locally is to use the following command:
 ```bash
-$ mdbook build
+$ mdbook serve
 ```
 
-The output will be in the `book` subdirectory. To check it out, open it in
-your web browser.
+This serves the book at http://localhost:3000, and rebuilds it on changes.
+You can now view the book in your web browser. If you make changes to the book's source code, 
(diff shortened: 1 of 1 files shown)


src/editions/transitioning-an-existing-project-to-a-new-edition.md
```diff
@@ -21,7 +21,7 @@ trait Foo {
 ```
 
 This code uses an anonymous parameter, that `Box<Foo>`. This is [not
-supported in Rust 2018](rust-2018/trait-system/no-anon-params.html), and
+supported in Rust 2018](/rust-2018/trait-system/no-anon-params.html), and
```
steveklabnik: we cannot use links like this, as this stuff has to work offline in the main distribution

synaestheory: How do you suggest I move forward?

Links on the website are broken because they are considered relative links without that leading slash. 

`mdbook serve` makes it available on your local machine even when you are "offline", but it sounds like I'm missing some other requirement that you might be aware of.

ehuss:
@synaestheory Links are now updated as of #152.  Would you be interested in rebasing this just with the README changes?

synaestheory:
Sorry I hadn't been able to get around to this sooner. I've rebased to only include the changes to the README.

PR #134, opened 2019-01-10
dyn-trait-for-trait-objects.md: Fix link to `impl Trait` explanation

xtian:


ehuss:
I don't think the link is broken.  The edition guide is published on the official site with mdbook 0.1 which uses different (absolute) links.  See https://doc.rust-lang.org/stable/edition-guide/rust-2018/trait-system/dyn-trait-for-trait-objects.html for how it looks.

@steveklabnik do you want help changing travis here to use mdbook 0.1.  Or, more ambitiously, I offer to help transition all of the official books to 0.2.

steveklabnik:
You're right, @ehuss !

So, @jamesmunns has https://github.com/rust-lang/rust/pull/56291 open, which would allow for per-book mdbook versions, and let us upgrade piecemeal. If he lands that, we can update the books one at a time, which I think is the most realistic option. James, do you need help getting that PR across the line? After we land that, help updating all of the books would be *very* appreciated, for sure.

jamesmunns:
Hey @steveklabnik, I haven't had time to dig into mdbook to correct the issue of rendering the `print.html` correctly, which is blocking that PR at the moment.

What I can do very quickly, is to remove the embedded portion of that PR, and only add the MdBook 1/2 compatibility section, and land that very soon. This would come with the warning that @ehuss 's changes might also trip over the same print.html issues we did, depending on how links are used (I haven't reviewed this PR yet).

If I do split this up, would you (Steve) prefer a new PR that only does the 1/2 compat, or would you like me to repurpose the old PR, and open a new one to add embedded later? The later part will probably require an upgrade of the mdbook2 dependency.

xtian:
Ah sorry, the published version I was using was at https://rust-lang-nursery.github.io/edition-guide/rust-2018/trait-system/dyn-trait-for-trait-objects.html where the link is broken. Should have noted that in the description. Is this intended?

Ppjet6:
#143 fixes this and a bunch of other links.

ehuss:
Thanks for the PR! The links should now be updated with #152.

PR #136, opened 2019-01-14
rust-2018/control-flow/index: Fix link

behnam:
The link in <https://rust-lang-nursery.github.io/edition-guide/rust-2018/control-flow/index.html> points to <https://rust-lang-nursery.github.io/edition-guide/rust-2018/control-flow/rust-2018/control-flow/async-await-for-easier-concurrency.html>, as the destination is relative, but the path is from the root. Remove the non-relative part of the path to fix the link.


Ppjet6:
#143 fixes this and a bunch of other links.

ehuss:
Thanks for the PR! The links should now be updated with #152.

PR #140, opened 2019-01-29
Fix edition-changes links

### LINKED ISSUE #138: All links broken on https://rust-lang-nursery.github.io/edition-guide/rust-2018/edition-changes.html
Every link here takes me to 404 - it creates URLS containing incorrect "rust-2018/rust-2018" paths.

This looks pretty bad and it's hopefully an easy fix.

dwijnand:
Fixes #138

### DIFF
--- a/src/rust-2018/edition-changes.md
+++ b/src/rust-2018/edition-changes.md
@@ -36,15 +36,15 @@ the 2018 edition compared to the 2015 edition.
   specify `cargo install --path .` to install the current package.
 
 [#55618]: https://github.com/rust-lang/rust/issues/55618
-[Anonymous trait function parameters]: rust-2018/trait-system/no-anon-params.html
-[At most once]: rust-2018/macros/at-most-once.html
-[Non-lexical lifetimes]: rust-2018/ownership-and-lifetimes/non-lexical-lifetimes.html
-[Path changes]: rust-2018/module-system/path-clarity.html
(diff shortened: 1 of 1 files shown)


dwijnand:
Tested with `mdbook build`.

steveklabnik:
The reason I haven't checked out / merged this yet is that I'm not sure if the issue here is mdbook 0.1 vs 0.2. we have to make sure everything works with 0.2 before we can update, and only a few hours ago was it even possible to use 0.2 with this at all. So, sorry for the wait, but that's the root issue here.

dwijnand:
I see.  I'm not sure why you say

> only a few hours ago was it even possible to use 0.2 with this at all

because I have mdbook v0.2.3 installed locally and that's what I used to verify this.  So I'm confused.

steveklabnik:
The edition guide is currently built with 0.1.

steveklabnik:
Sorry, even more context:

There are changes between how links work between mdbook 0.1 and 0.2. So, I believe that what exists right now works with 0.1, but since you built with 0.2, they seemed broken. 

Upstream builds with 0.1. This updates to the 0.2 style, which works locally but would fail upstream.

 Upstream got the ability to build with 0.2 last night.

Does that make sense?

dwijnand:
Yes, thank you, Steve.  Sorry for stealing focus.

This patch is trivial, I'll assume it'll be subsumed by the upcoming PR to upgrade the edition guide to mdbook 0.2.

PR #141, opened 2019-01-30
Update for uniform_path stabilization.

### LINKED ISSUE #139: Seemingly incorrect info about path clarity
In "2018-specific changes":

- "use declarations must begin with ... an external crate name"

This does not seem to be true - paths like "use foo::bar" will resolve a "foo" in local scope, not just a crate name.

- "to catch possible ambiguities with uniform paths (which is planned for inclusion soon #55618)."

That issue is closed and presumably in stable?

In "Path clarity":

- "Absolute paths begin with a crate name, where the keyword crate refers to the current crate."

It's not clear what "absolute paths" means. Is this for "::"? It should be explicit.

Again on this page I don't see mention of unprefixed paths resolving names in local scope, which they do.

- "In Rust 2018, paths in use declarations must begin with a crate name, crate, self, or super"

Again, 'use' declarations seem to resolve names in local scope. No mention of "::" here.

All of the above applies with or with out the 'uniform_paths' feature.

ehuss:
This is just a rough stab at improving things, or at least removing some of the wrong information.  I think more details will need to wait for the reference.

I rewrote `Changes to paths` because it no longer matched what current 2018 code should look like, and had a lot of overlap with what "uniform paths" was saying.  I replaced it with something roughly describing extern prelude, which I think fits?

Closes #139.

### DIFF
--- a/src/rust-2018/module-system/path-clarity.md
+++ b/src/rust-2018/module-system/path-clarity.md
@@ -20,7 +20,7 @@ Here's a brief summary:
 * Paths starting with `::` must reference an external crate.
 * A `foo.rs` and `foo/` subdirectory may coexist; `mod.rs` is no longer needed
   when placing submodules in a subdirectory.
-* `use` declarations take [uniform paths](#uniform-paths).
+* Paths in `use` declarations work the same as other paths.
 
 These may seem like arbitrary new rules when put this way, but the mental
 model is now significantly simplified overall. Read on for more details!
@@ -140,9 +140,6 @@ mod submodule {
         // but in a submodule it requires a leading :: if not imported with `use`
         let x = ::chrono::Utc::now();
     }
-
-    // unlike expressions, `use` paths were allowed to reference crates directly
-    use chrono::Local;
 }
 ```
 
@@ -162,9 +159,6 @@ mod submodule {
         // crates may be referenced directly, even in submodules
         let x = chrono::Utc::now();
     }
-
-    // `use` paths have the same path style
-    use chrono::Local;
 }
 ```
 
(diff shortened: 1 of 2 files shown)


src/rust-2018/module-system/path-clarity.md
```diff
@@ -268,21 +203,21 @@ see their names, instead of having a bunch of tabs named `mod.rs`.
 
 # Uniform paths
 
-> Uniform paths are a nightly-only feature.
+![Minimum Rust version: 1.32](https://img.shields.io/badge/Minimum%20Rust%20Version-1.32-brightgreen.svg)
 
 The uniform paths variant of Rust 2018 simplifies and unifies path handling
```
Centril: Not part of your diffs, but we should probably not talk about a "variant" anymore since the anchored paths proposal was rejected.

src/rust-2018/module-system/path-clarity.md
```diff
@@ -268,21 +203,21 @@ see their names, instead of having a bunch of tabs named `mod.rs`.
...
 compared to Rust 2015. In Rust 2015, paths work differently in `use`
 declarations than they do elsewhere. In particular, paths in `use`
 declarations would always start from the crate root, while paths in other code
-implicitly started from the current module. Those differences didn't have any
+implicitly started from the current scope. Those differences didn't have any
 effect in the top-level module, which meant that everything would seem
 straightforward until working on a project large enough to have submodules.
 
 In the uniform paths variant of Rust 2018, paths in `use` declarations and in
```
Centril: Same here re. "variant"...

src/rust-2018/module-system/path-clarity.md
```diff
@@ -268,21 +203,21 @@ see their names, instead of having a bunch of tabs named `mod.rs`.
...
-other code always work the same way, both in the top-level module and in any
-submodule. You can always use a relative path from the current module, a path
-starting from an external crate name, or a path starting with `crate`, `super`,
-or `self`.
+other code almost always work the same way, both in the top-level module and
```
Centril: The "almost" bit here will leave the reader wondering...

Centril: Maybe: "paths in `use` declarations and in other code work the same way, ...".

This is less "100% sure" but also doesn't say "maybe".

johnthagen:
FYI, https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html

Should this badge get updated?

screenshot

ehuss:
@johnthagen this PR updates the badge.

Ping @Centril or @steveklabnik is there anything else you'd like to see for this PR?

Centril:
@ehuss I think I just forgot about the PR... sorry :)

PR #143, opened 2019-02-05
Fix breaking internal links

h-michael:
related with #138, #140


ehuss:
Thanks for the PR!  The links should now be updated with #152.

PR #144, opened 2019-02-10
Add example to rustup to show overriding to specific version

senorsmile:


src/rust-2018/rustup-for-managing-rust-versions.md
```diff
@@ -209,4 +214,4 @@ $ rustup component add llvm-tools-preview
 ```
 
 This is the newest component, and so doesn't have good documentation at the
-moment.
\ No newline at end of file
+moment.
```
senorsmile: There is no change on this line.  Any idea why it's showing that?

steveklabnik: You removed a space at the end of the line, which is good.

senorsmile: Interesting.  Vim must have done that automatically with one of the various plugins I've installed.

PR #145, opened 2019-02-12
Fix links to other pages of the book in rust-2018/

Ppjet6:
It seems these urls contain two `rust-2018/`, they're being appended to the current url which already has `rust-2018/` in it.

Signed-off-by: Maxime “pep” Buquet <pep@bouah.net>


Ppjet6:
#143 actually already fixes this, closing :)

PR #147, opened 2019-02-27
Fix issue with rust's linkchecker and mdbook.

ehuss:
The linkchecker in the rust repo rejects this because there are duplicate
HTML id's. The theme selector uses the id "rust" for the Rust color theme.
See https://github.com/rust-lang-nursery/mdBook/issues/880 for more info.

This blocks updating upstream.


Centril:
I gave you write access to the repo since you've been contributing a lot. Feel free to merge when you deem appropriate.

Issue #76, opened 2018-08-06
Move the edition guide to a more prominent location


Centril:
We should eventually move the guide out of the https://rust-lang-nursery.github.io/ URL and give it a more official-looking, short and sweet URL instead.

We decided that we will punt on this for Preview 2; but we should do this for Release Candidate 1.

A suggested URL for the whole edition is https://rust-lang.org/edition/ and https://rust-lang.org/edition/2018/  then goes to the 2018 page.

We can also move it to  https://docs.rust-lang.org/edition/

orium:
Note that when this is done some urls in cargo will need to be updated.  At the time of this comment there are two references to `rust-lang-nursery.github.io`:

```text
orium@laptop-orion cargo % rp rust-lang-nursery.github.io
src/cargo/util/diagnostic_server.rs
174:  https://rust-lang-nursery.github.io/edition-guide/editions/transitioning-your-code-to-a-new-edition.html
196:  https://rust-lang-nursery.github.io/edition-guide/editions/transitioning-your-code-to-a-new-edition.html
```

steveklabnik:
This is not happening as part of the release candidate, so de-milestoning.

tmandry:
Will this happen before release? It looks like there was some mention on Discord, but no formal decision made.

At least one rustc PR (https://github.com/rust-lang/rust/pull/55185) includes a guide URL; we're not sure if it's going to change.

dwijnand:
https://doc.rust-lang.org/stable/edition-guide/ is now a thing, but note that it's some other version of the guide that isn't https://rust-lang-nursery.github.io/edition-guide/.

Note the "Minimum Rust Version" in:
* https://doc.rust-lang.org/stable/edition-guide/rust-2018/ownership-and-lifetimes/non-lexical-lifetimes.html, vs
* https://rust-lang-nursery.github.io/edition-guide/rust-2018/ownership-and-lifetimes/non-lexical-lifetimes.html

dwijnand:
Interestingly hasn't changed with the 1.32 release.

siddharthasahu:
I think the minimum versions needs an update at https://doc.rust-lang.org/stable/edition-guide/rust-2018/documentation/new-editions-of-the-book.html

ehuss:
The guide now lives at https://doc.rust-lang.org/edition-guide/.  The latest version will be published at https://doc.rust-lang.org/nightly/edition-guide/, and should be live soon.

[...]

PR #150, opened 2019-02-28
Redirect gh-pages to doc.rust-lang.org.

ehuss:
This redirects to the nightly copy with the intent that people were coming here
for the up-to-date version. This includes fragment-preserving redirects.

cc #76, #138


Centril:
(merge this when you are happy... ^^)

PR #158, opened 2019-05-06
Fix typo in controlling-panics-with-std-panic.md

jplatte:
I copied the example and was hit with a compiler warning.


jplatte:
Apparently the GitHub web editor changed a line ending. I'll fix that later if it's a problem.

ehuss:
Thanks!

PR #160, opened 2019-05-20
Fix error example for modern compilers.

### LINKED ISSUE #159: non-lexical lifetimes
https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/improved-error-messages.html
The code in this example will compile, because of non-lexical lifetimes, described previously.
https://doc.rust-lang.org/edition-guide/rust-2018/ownership-and-lifetimes/non-lexical-lifetimes.html
Consider this example

```
fn main() {
    let mut x = 5;
    let y = &x;

    x += 1;
    println!("{}",*y);
}
```

ehuss:
NLL makes the example not an error anymore, but a small tweak makes it fail again.

Fixes #159.

### DIFF
--- a/src/rust-2018/the-compiler/improved-error-messages.md
+++ b/src/rust-2018/the-compiler/improved-error-messages.md
@@ -8,41 +8,42 @@ error message system was created.
 
 For example, here's some code that produces an error:
 
-```rust,ignore
+```rust,compile_fail
 fn main() {
     let mut x = 5;
-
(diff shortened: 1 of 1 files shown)


ehuss:
@Centril Do you think you can give this a quick review?  I think it is a fairly straightforward update to ensure the example still fails with NLL.

Centril:
Something about this PR makes me 😄.

Issue #184, opened 2019-07-13
Some confusion by reading about `install-update -a` command.


dkotrada:
I am reading [this page](https://doc.rust-lang.org/stable/edition-guide/rust-2018/cargo-and-crates-io/cargo-install-for-easy-installation-of-tools.html) now.

At the bottom there is `cargo install-update -a` command suggestion.

I don't understand why is there `Minimum Rust Version 1.5` badge? Rust nightly is 1.38 at the moment.

I am not native speaker and it is a little bit confusing to me. I did not seen this sentence `This will allow you to use this command`

For me it were better like that:

> This will allow you to use `cargo install-update -a`command ...

dkotrada:
If it is ok dkotrada:patch-1

PR #185, opened 2019-07-13
#184 More clear explanation and Title.

dkotrada:
Since a new thought is introduced here. It is better to separate it by a corresponding title.


ehuss:
Thanks!

PR #187, opened 2019-07-21
Attempt to clarify "no more mod.rs".

ehuss:
The example code was a little confusing. Instead, use a table which compares
the directory structure.

Not sure if this is clear enough, it's a little awkward to explain.

[Rendered](https://github.com/ehuss/edition-guide/blob/fb53ec9b6915760013865e62772f3404b3563e08/src/rust-2018/module-system/path-clarity.md#no-more-modrs)


Centril:
Much better!

PR #194, opened 2019-12-15
Updated install and uninstall instructions

zakaluka:
rust-lang/rustup#2148

Signed-off-by: zakaluka <zakthrow@gmail.com>

### DIFF
--- a/src/rust-2018/rustup-for-managing-rust-versions.md
+++ b/src/rust-2018/rustup-for-managing-rust-versions.md
@@ -14,24 +14,24 @@ To install Rust through Rustup, you can go to
 so on your platform. This will install both `rustup` itself and the `stable`
 version of `rustc` and `cargo`.
 
-To install a specific Rust version, you can use `rustup install`:
+To install a specific Rust version, you can use `rustup toolchain install`:
 
 ```console
-$ rustup install 1.30.0
+$ rustup toolchain install 1.30.0
 ```
 
 This works for a specific nightly, as well:
 
 ```console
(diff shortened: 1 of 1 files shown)


zakaluka:
As part of the update, I also added two sentences in the edition book for how to uninstall a specific version of Rust.

ehuss:
@kinnison I'd encourage reconsidering deprecating or hiding the alias.  I think the feedback in https://github.com/rust-lang/rustup/issues/2148 and related threads is clear that there is a general objection by users to the change.  I'm still unclear as to why it needs to be done in the first place.  I think there have been several proposals (including mine) to reduce the complexity or maintenance burden of having an alias.  I see you mentioned migrating some functionality to Cargo, but there is no plan or intent to do that, that I am aware of (there has been talk about it in the past, but nothing concrete).

This PR itself probably doesn't really matter (I doubt this page receives much traffic).  The commands seem more complex to my eyes.  But since they are functionally identical, and I don't want to spend more time on it, I'll go ahead and merge it.

kinnison:
@ehuss The alias has never not been hidden, but we're definitely intending to support the funtionality into the future once the appropriate capabilities are in clap to ensure they can never get out of sync again.  Thank you for that merge.  I was definitely convinced, by you and others, that *removing* the support would be a bad thing, but the short form has never been visible in Rustup's CLI by default and to my mind was at least slightly confusing given that rustup manages toolchains, components, and targets, all with similar CLI APIs.  I won't be pushing hard on this, but I am grateful for the merge nonetheless.

PR #197, opened 2020-02-21
Use rust-lang/rust linkchecker on CI.

ehuss:
This can help ensure submodule updates go smoothly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,11 +13,11 @@ jobs:
       run: |
         rustup set profile minimal
         rustup toolchain install nightly -c rust-docs
-        rustup default nightly 
+        rustup default nightly
     - name: Install mdbook
       run: |
         mkdir bin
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -11,18 +11,27 @@ jobs:
       run: rustup self update
     - name: Install Rust
       run: |
+        set -e
```
pietroalbini: This is not needed on GitHub Actions (it's already the default).

ehuss: ah, I had tested azure and travis, and I was assuming gha was the same as azure. thanks!  (trying to update everyone, and everyone uses different CI)

ehuss: ah, I had tested azure and travis, and I was assuming gha was the same as azure. thanks! (trying to update everyone, and everyone uses different CI)

PR #213, opened 2020-05-11
Changes for Rust 1.32 & setup for edition-next

### LINKED ISSUE #201: Update guide for Rust 1.32
We have not yet added any features from Rust 1.32; I will go back, look at the release, and do so.

steveklabnik:
This change adds in things stabilized in Rust 1.32. However, there's a few things that are a bit of a bigger deal than that.

Fixes #201. 

For starters, this begins a section for a new edition. Since we have not decided if and when one is happening yet, I've called this "Edition next", because "next" has four characters, so it's kind of similar to a year, which has four numbers.

Within that, I've just added sections. I don't think we can do a proper grouping until we've gotten more stuff.

I believe some of this stuff is Rust 2018 specific, so I figure that it should probably be added to [2018-specific changes](https://doc.rust-lang.org/stable/edition-guide/rust-2018/edition-changes.html), but did not do that yet, pending us being settled on all the previous stuff.


aidanhs:
:+1:

PR #218, opened 2020-08-27
Fix a small typo.

teymour-aldridge:


teymour-aldridge:
Another possible mistake is that this file says "You can use assignment and assignment operators:" (which sounds wrong to me)

ehuss:
> Another possible mistake is that this file says "You can use assignment and assignment operators:" (which sounds wrong to me)

What sounds wrong about it?  Assignment expressions are allowed in const contexts.  More information is [here](https://doc.rust-lang.org/nightly/reference/const_eval.html#constant-expressions).

teymour-aldridge:
Oh sorry, I misread – I thought it said you can use "assignment and assingment".
Missed the bit about "operators"!

PR #219, opened 2020-08-29
Clarify the limitation of ? in main and tests

davidOSUL:
I was trying to figure out why my error messages were not printing how I wanted them to (using Display). This page in the edition book is one of the first links that pops up when I googled it, but it wasn't till I saw [this post on the rust forum](https://users.rust-lang.org/t/how-to-customize-the-main-error-message-when-returning-an-error/21903) that I realized there was no way to do this "automatically", so I thought it might be helpful for others to clarify in these docs as well.


ehuss (approved):
Thanks!

PR #223, opened 2020-11-02
Clarify slice patterns example

tiziano88:
The previous phrasing was not precise enough, and suggested that the `[first, second]` pattern would match any slice with at least two elements, while in fact it only matches slices with exactly two elements.


steveklabnik (approved):
Thanks!

Issue #178, opened 2019-06-14
Tracking issue for wasm-unknown-unknown


carols10cents:
Not sure about Rust issue, maybe https://github.com/rust-lang/rust/issues/44839 ?

Places that need to be updated:
- The nightly version at the top: https://github.com/rust-lang-nursery/edition-guide/blame/68c458b9b68edced39aebea60e672a22d4d50edf/src/rust-2018/platform-and-target-support/webassembly-support.md#L5
- This spot: https://github.com/rust-lang-nursery/edition-guide/blame/68c458b9b68edced39aebea60e672a22d4d50edf/src/rust-2018/platform-and-target-support/webassembly-support.md#L24-L28

> However, in the meantime, Rust has also grown its own support, independent from Emscripten. This is known as "the unknown target", because instead of `wasm32-unknown-emscripten`, it's `wasm32-unknown-unknown`. This will be the preferred target to use once it's ready, but for now, it's really only well-supported in nightly.

ehuss:
Closing per #232, this section has been removed.  Indeed there isn't great documentation for targets.  Ideally I think that should live in the `rustc` book, and is something that has been on my todo list to get started for a while.

PR #224, opened 2020-11-19
Update notes about static linking and MUSL

joshtriplett:
People often use the edition guide as general documentation of
capabilities, not just for Rust 2018. There is extensive information
on the web suggesting that Rust *only* supports static linking with
MUSL. Add a note documenting current capabilities, so that people know
they have a choice of libc implementations even if they need static
linking.

### DIFF
--- a/src/rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.md
+++ b/src/rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.md
@@ -2,6 +2,9 @@
 
 ![Minimum Rust version: 1.1](https://img.shields.io/badge/Minimum%20Rust%20Version-1.1-brightgreen.svg)
 
+(Note: current Rust (as of 1.48) supports static linking with either glibc or
+MUSL, so you can select whichever libc you need or prefer.)
+
 By default, Rust will statically link all Rust code. However, if you use the
 standard library, it will dynamically link to the system's `libc`
 implementation.


joshtriplett:
cc @estebank

steveklabnik:
Thanks @joshtriplett ! Just so you know, the current state of the edition RFC would be to remove the vast majority of the edition guide, and until that RFC is merged, I've effectively stopped work here entirely.

joshtriplett:
@steveklabnik That's fair, but until that change happens, I've seen people use and cite the edition guide as a reference for the capabilities of Rust. I don't mind if this text ends up deleted later to make way for something else. Would you consider merging it in the interim?

Mark-Simulacrum:
I think we shouldn't just remove the information -- I would like to see it put elsewhere (maybe something like a "Feature List" or whatever, not sure on naming), but regardless I agree that merging this in the interim makes sense to me. The reason we are thinking of dropping it from the Edition Guide in particular is that people were confused about whether the features were tied to a particular edition or not, I believe.

src/rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.md
```diff
@@ -2,6 +2,9 @@
 
 ![Minimum Rust version: 1.1](https://img.shields.io/badge/Minimum%20Rust%20Version-1.1-brightgreen.svg)
 
+(Note: current Rust (as of 1.48) supports static linking with either glibc or
+MUSL, so you can select whichever libc you need or prefer.)
```
ehuss, line 6: Are you sure it is 1.48?  I thought static support is in 1.49 (https://github.com/rust-lang/rust/pull/77386).

I would suggest being a little clearer here, and providing an example:

```suggestion
> Note: Starting with Rust 1.49, the Linux glibc targets also support static
> linking the C runtime. This can be done with the [`+crt-static` target feature][crt-static].
> For example, with Cargo:
>
> ```console
> RUSTFLAGS="-C target-feature=+crt-static" cargo build --target x86_64-unknown-linux-gnu
> ```

[crt-static]: ../../../reference/linkage.html#static-and-dynamic-c-runtimes
```

I would also put this at the bottom of this section (just before `## Installing MUSL support`).

ehuss:
I'm fine with minor updates until the fate of the guide is determined.

Also, in the future I would like to see platform-specific documentation added to the `rustc` book.  I would like to see the https://doc.rust-lang.org/nightly/rustc/targets/index.html chapter include sub-chapters that go into detail on how to use each target, such as dependencies to install, peculiarities, etc.

ehuss:
Ping @joshtriplett Are you still interested in this?

ehuss:
Closing as this section has been removed in #232.

Similar to #178, ideally I think the rustc book should have documentation on the peculiarities of different targets. It has long been on my todo list, but alas never started.  Hopefully at some point this information can be moved over there.

PR #226, opened 2020-12-07
Update code generated by `cargo fix --edition`

ejez:
Update the code generated by `cargo fix --edition` in the given example to include `dyn` in the trait object function parameter.


ejez:
While continuing to read the page, I found more outdated info regarding `cargo fix`, (for example, the flag `--edition-idioms` being needed to get `dyn` in the generated code, while now using only `cargo fix --edition` inserts it).

This page needs an update in different parts (maybe changing the provided example), so now closing this PR, and instead will open an issue.

PR #230, opened 2021-02-02
Update musl libc link

epompeii:
musl libc has moved to from https://www.musl-libc.org/ to https://musl.libc.org/


ehuss (approved):
Thanks!

PR #232, opened 2021-05-03
Minimize the edition guide

nikomatsakis:
This lays out the new, more minimal structure of the edition guide. I copied and pasted the existing content that was still relevant and also copied in the introduction and the descriptions of the changes from @m-ou-se's blog. 

cc @rust-lang/project-edition-2021


src/rust-2021/reserving_syntax.md
```diff
@@ -0,0 +1,40 @@
...
+
+- `f""` as a short-hand for a format string.
+  For example, `f"hello {name}"` as a short-hand for the equivalent `format_args!()` invocation.
+
+- `c""` or `z""` for null-terminated C strings.
+
+- `k#keyword` to allow writing keywords that don't exist yet in the current edition.
+  For example, while `async` is not a keyword in edition 2015,
+  this prefix would've allowed us to accept `k#async` in edition 2015
+  without having to wait for edition 2018 to reserve `async` as a keyword.
```
skippy10110, line 38: Changing "without having to wait for" to "while we wait for" might make it clearer that async still becomes a keyword in 2018, even with k# available. (I think that's the intent, if this had been available then)

m-ou-se: Thanks! Updated in https://github.com/rust-lang/blog.rust-lang.org/pull/825.

src/rust-2018/index.md
```diff
@@ -1,8 +1,9 @@
...
-Rust 2018 smooths off rough edges, makes writing code simpler and easier,
-and removes some inconsistencies.
\ No newline at end of file
+| Info | |
+| --- | --- |
+| RFC | [#2052](https://rust-lang.github.io/rfcs/2052-epochs.html), which also proposed the Edition system |
+| Release version | [1.31.0](https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html) |
+
+The edition system was created for the release of Rust 2018. The release of the Rust 2018 edition coincided with a number of other features all coordinated around the theme of *productivity*. The majority of those features were backwards compatible and are now available on all editions; however, some of those changes required the edition mechanism (most notably the [module system changes](path-changes.md)).
+s
```
ehuss, line 9: Stray `s`

ehuss (commented):
A few high-level comments:

- I would prefer to keep the filename style using hyphens instead of underscores, to be consistent with the rest of the books.
- I think it would be good to add redirects for the removed pages, otherwise there will be a lot of broken links.  mdbook now supports that in book.toml.  I *think* that should work, though the rust-lang linkchecker doesn't work with it, I don't think there are any relative links to the edition guide.

It also looks like there are a few errors and broken links here. Let me know if you want help with any of the above.

nikomatsakis:
@ehuss 

> I would prefer to keep the filename style using hyphens instead of underscores, to be consistent with the rest of the books.

OK! No opinion here. 

I do have a question: this book uses `index.md` files, is this something we do elsewhere?

> I think it would be good to add redirects for the removed pages, otherwise there will be a lot of broken links. mdbook now supports that in book.toml. I think that should work, though the rust-lang linkchecker doesn't work well with it.

Good point. I'll look into what support mdbook has for that.

ehuss:
> I do have a question: this book uses `index.md` files, is this something we do elsewhere? (vs `src/rust-2015.md`, for example)

I think it is fine to use `index.md`.  It is not used commonly elsewhere, but I think that's just an artifact of evolution. For example, the reference had `src/types.md` as a single file. When it was split into multiple files, I kept the original `src/types.md` instead of renaming it to `src/types/index.md` because that would break old links and was too much of a hassle to deal with (and mdbook didn't support renames at the time).  Books like TRPL use a flat structure (no directories), though I'm not sure why.

nikomatsakis:
I guess modifying it would require modifying links, and I have no desire to do *that*.

steveklabnik:
>  Books like TRPL use a flat structure (no directories), though I'm not sure why.

Because the book is older than mdbook is, and the initial build infrastructure was *very* simple.

nikomatsakis:
Addressed @ehuss's points:

> I would prefer to keep the filename style using hyphens instead of underscores, to be consistent with the rest of the books.

Done!

> I think it would be good to add redirects for the removed pages, otherwise there will be a lot of broken links. mdbook now supports that in book.toml. I think that should work, though the rust-lang linkchecker doesn't work well with it.

Done! Since most of that content no longer exists, I tried to map to the "closest relevant thing".

> It also looks like there are a few errors and broken links here. Let me know if you want help with any of the above.

mdbook test passes now, if that's what you mean, although I mostly just ignored code examples. Maybe as we revise those chapters we can convert them into tests that actually run.

nikomatsakis:
I think this is ready to land, then.

nikomatsakis:
Oh, wait, I forgot about the rust-next pages.

ehuss:
I pushed a commit to fix some of the links here: https://github.com/ehuss/edition-guide/commit/8b6d2d1a22b96512b88958915ff656c48f318f1f

I can push it to your branch, or you can maybe copy it if you want.

The CI was a little out of date.  Links can't be absolute, since they would break when viewed from the filesystem.  Also, I added redirects to pages that might be more relevant to the original content.

nikomatsakis:
@ehuss feel free to push to my branch!

nikomatsakis:
Thanks @ehuss! I'm going to go ahead and merge this then.

PR #246, opened 2021-07-01
Provide more information about disjoint capture in closures and migration instructions

roxelo:
This PR adds additional information about disjoint capture in closures such as changes in semantics and migration instructions.

r? @nikomatsakis

Closes https://github.com/rust-lang/edition-guide/issues/241

### DIFF
--- a/src/rust-2021/disjoint-capture-in-closures.md
+++ b/src/rust-2021/disjoint-capture-in-closures.md
@@ -3,7 +3,7 @@
 ## Summary
 
 - `|| a.x + 1` now captures only `a.x` instead of `a`.
-- This can subtly change the drop order of things.
+- This can subtly change the drop order of things, and auto traits usage with closures.
 
 ## Details
 
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
 c();
 ```
 
-Starting in Rust 2021, closures will only capture the fields that they use.
+Starting in Rust 2021, closures will only capture the fields that they use, eliminating common borrow check errors.
 So, the above example will compile fine in Rust 2021.
 
-This new behavior is only activated in the new edition,
-since it can change the order in which fields are dropped.
-As for all edition changes, an automatic migration is available,
-which will update your closures for which this matters.
-It can insert `let _ = &a;` inside the closure to force the entire
-struct to be captured as before.
+Disjoint capture was proposed as part of [RFC 2229](https://github.com/rust-lang/rfcs/blob/master/text/2229-capture-disjoint-fields.md), and we suggest reading the RFC to better understand motivations behind the feature.
+
+## Changes to semantics because of disjoint capture
+`#[feature(capture_disjoint_fields)]` introduces (minor) breaking change to the language. This means that there might be observable changes or valid Rust 2018 code that fails to compile once you move to Rust Edition 2021. You can use `cargo fix` with the `disjoint_capture_migrations` lint to migrate your existing code to Rust 2021.
+### Wild Card Patterns
+Closures now only capture data that needs to be read, which means the following closures will not capture `x`
+
+```rust,ignore
+let x = 10;
+let c = || {
+    let _ = x;
+};
+
+let c = || match x {
+    _ => prinln!("Hello World!")
+};
+```
+
+### Drop Order
+Since only a part of a variable might be captured instead of the entire variable, when different fields or elements (in case of tuple) get drop, the drop order might be affected.
+
+```rust,ignore
+{
+    let t = (vec![0], vec![0]);
+
+    {
+        let c = || {
+            move(t.0);
+        };
+    } // c and t.0 dropped here
+} // t.1 dropped here
+
+```
+
+
+### Auto Traits
+Structs or tuples that implement (an) auto trait(s) and are passed along in a closure may no longer guarantee that the closure can also implement that/those auto trait(s).
+
+For instance, a common way to allow passing around raw pointers between threads is to wrap them in a struct and then implement `Send`/`Sync` auto trait for the wrapper. The closure that is passed to `thread::spawn` uses the specific fields within the wrapper but the entire wrapper is captured regardless. Since the wrapper is `Send`/`Sync`, the code is considered safe and therefore compiles successfully.
+
+With this feature only the specific field mentioned in the closure gets captured, which wasn't originally `Send`/`Sync` defeating the purpose of the wrapper.
+
+
+```rust,ignore
+struct Ptr(*mut i32);
+unsafe impl Send for Ptr;
+
+
+let mut x = 5;
+let px = (&mut x as *mut i32);
+
+let c = thread::spawn(move || {
+    *(px.0) += 10;
+}); // Closure captured px.0 which is not Send
+```
+
+## Migrations
+
+This new behavior is only activated in the new edition, or by specifically enabling it using `#[feature(capture_disjoint_fields)]`,
+since it can change the order in which fields are dropped and can impact auto trait usage with closures.
+
+As for all edition changes, an automatic migration is available.
+If you would like to be warned of semantics change that may impact your code, you can [use the lint](https://doc.rust-lang.org/rustc/lints/levels.html) `disjoint_capture_migrations`. The lint is also supported with [cargo fix](https://doc.rust-lang.org/cargo/commands/cargo-fix.html) to automatically migrate your code.
+
+The migration fix involves adding `let _ = &a;` inside the closure to force the entire variable to be captured as before.
--- a/src/rust-2021/disjoint-capture-in-closures.md
+++ b/src/rust-2021/disjoint-capture-in-closures.md
@@ -35,8 +35,6 @@ Changing the variables captured by a closure can cause programs to change behavi
 - changes to drop order, or when destructors run ([details](#drop-order));
 - changes to which traits a closure implements ([details](#trait-implementations)).
 
-When running `cargo fix --edition`, Cargo will update the closures in your code to help you migrate to Rust 2021, as described below in [Migrations](#migrations).
-
 Whenever any of the scenarios below are detected, `cargo fix` will insert a "dummy let" into your closure to force it to capture the entire variable:
 
 ```rust
--- a/src/rust-2021/disjoint-capture-in-closures.md
+++ b/src/rust-2021/disjoint-capture-in-closures.md
@@ -3,7 +3,7 @@
 ## Summary
 
 - `|| a.x + 1` now captures only `a.x` instead of `a`.
-- This can subtly change the drop order of things, and auto traits usage with closures.
+- This can subtly change the drop order of things and whether auto traits are applied to closures or not.
 
 ## Details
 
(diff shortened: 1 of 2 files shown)


src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
-This new behavior is only activated in the new edition,
-since it can change the order in which fields are dropped.
-As for all edition changes, an automatic migration is available,
-which will update your closures for which this matters.
-It can insert `let _ = &a;` inside the closure to force the entire
-struct to be captured as before.
+Disjoint capture was proposed as part of [RFC 2229](https://github.com/rust-lang/rfcs/blob/master/text/2229-capture-disjoint-fields.md), and we suggest reading the RFC to better understand motivations behind the feature.
+
+## Changes to semantics because of disjoint capture
+`#[feature(capture_disjoint_fields)]` introduces (minor) breaking change to the language. This means that there might be observable changes or valid Rust 2018 code that fails to compile once you move to Rust Edition 2021. You can use `cargo fix` with the `disjoint_capture_migrations` lint to migrate your existing code to Rust 2021.
```
ehuss, line 38: I think it would be good to avoid discussing nightly-only controls, as the target audience here is someone running the stable 2021 release.

I also don't think there is a need to put much emphasis on the specific lint used to migrate, as the user doesn't really need to know about it or specify it.

```suggestion
Disjoint captures introduces a minor breaking change to the language. This means that there might be observable changes or valid Rust 2018 code that fails to compile once you move to Rust Edition 2021. 

When running `cargo fix --edition`, Cargo will update the closures in your code to help you migrate to Rust 2021, as described below in [Migrations](#migrations).
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+Closures now only capture data that needs to be read, which means the following closures will not capture `x`
+
+```rust,ignore
+let x = 10;
+let c = || {
+    let _ = x;
+};
+
+let c = || match x {
+    _ => prinln!("Hello World!")
```
ehuss, line 49:
```suggestion
    _ => println!("Hello World!")
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
-It can insert `let _ = &a;` inside the closure to force the entire
-struct to be captured as before.
+Disjoint capture was proposed as part of [RFC 2229](https://github.com/rust-lang/rfcs/blob/master/text/2229-capture-disjoint-fields.md), and we suggest reading the RFC to better understand motivations behind the feature.
+
+## Changes to semantics because of disjoint capture
+`#[feature(capture_disjoint_fields)]` introduces (minor) breaking change to the language. This means that there might be observable changes or valid Rust 2018 code that fails to compile once you move to Rust Edition 2021. You can use `cargo fix` with the `disjoint_capture_migrations` lint to migrate your existing code to Rust 2021.
+### Wild Card Patterns
+Closures now only capture data that needs to be read, which means the following closures will not capture `x`
+
+```rust,ignore
```
ehuss, line 42: I feel pretty uncomfortable using `ignore` in examples.  Would it be possible to make all of them testable?

```suggestion
```rust
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+            move(t.0);
+        };
+    } // c and t.0 dropped here
+} // t.1 dropped here
+
+```
+
+
+### Auto Traits
+Structs or tuples that implement (an) auto trait(s) and are passed along in a closure may no longer guarantee that the closure can also implement that/those auto trait(s).
```
ehuss, line 71:
```suggestion
Structs or tuples that implement auto traits and are passed along in a closure may no longer guarantee that the closure can also implement those auto traits.
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+
+For instance, a common way to allow passing around raw pointers between threads is to wrap them in a struct and then implement `Send`/`Sync` auto trait for the wrapper. The closure that is passed to `thread::spawn` uses the specific fields within the wrapper but the entire wrapper is captured regardless. Since the wrapper is `Send`/`Sync`, the code is considered safe and therefore compiles successfully.
+
+With this feature only the specific field mentioned in the closure gets captured, which wasn't originally `Send`/`Sync` defeating the purpose of the wrapper.
```
ehuss, line 75:
```suggestion
With disjoint captures, only the specific field mentioned in the closure gets captured, which wasn't originally `Send`/`Sync` defeating the purpose of the wrapper.
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+
+
+```rust,ignore
+struct Ptr(*mut i32);
+unsafe impl Send for Ptr;
```
ehuss, line 80:
```suggestion
unsafe impl Send for Ptr {}
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+
+
+let mut x = 5;
+let px = (&mut x as *mut i32);
```
ehuss, line 84:
```suggestion
let px = Ptr(&mut x as *mut i32);
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+
+let c = thread::spawn(move || {
+    *(px.0) += 10;
+}); // Closure captured px.0 which is not Send
+```
+
+## Migrations
+
+This new behavior is only activated in the new edition, or by specifically enabling it using `#[feature(capture_disjoint_fields)]`,
```
ehuss, line 93:
```suggestion
This new behavior is only activated starting in the 2021 edition,
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+since it can change the order in which fields are dropped and can impact auto trait usage with closures.
+
+As for all edition changes, an automatic migration is available.
+If you would like to be warned of semantics change that may impact your code, you can [use the lint](https://doc.rust-lang.org/rustc/lints/levels.html) `disjoint_capture_migrations`. The lint is also supported with [cargo fix](https://doc.rust-lang.org/cargo/commands/cargo-fix.html) to automatically migrate your code.
+
+The migration fix involves adding `let _ = &a;` inside the closure to force the entire variable to be captured as before.
```
ehuss, line 99: Note that the lint name is wrong here, but also there is discussion in https://github.com/rust-lang/rust/issues/85894 to rename it anyways.

```suggestion
When running `cargo fix --edition`, the closures in your code may be updated so that they will retain the old behavior on both the 2018 and 2021 editions.
It does this by enabling the `disjoint_capture_migration` lint which adds statements like `let _ = &a;` inside the closure to force the entire variable to be captured as before.

After migrating, it is recommended to inspect the changes and see if they are necessary.
After changing the edition to 2021, you can try to remove the new statements and test that the closure works as expected.
You should review these closures, and ensure that the changes described above will not cause any problems.
```

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+
+
+let mut x = 5;
+let px = (&mut x as *mut i32);
+
+let c = thread::spawn(move || {
+    *(px.0) += 10;
+}); // Closure captured px.0 which is not Send
+```
+
```
ehuss, line 90: I feel like it would be nice to have a discussion in these sections about what the idiomatic code would look like in 2021.  Are people expected to write `let _ = &px;` under normal circumstances?  Or is there another way this example could be cleanly written in 2021?

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -29,12 +29,71 @@ let c = || println!("{}", a.y); // Error: Tries to capture all of `a`
...
+### Drop Order
+Since only a part of a variable might be captured instead of the entire variable, when different fields or elements (in case of tuple) get drop, the drop order might be affected.
+
+```rust,ignore
+{
+    let t = (vec![0], vec![0]);
+
+    {
+        let c = || {
+            move(t.0);
```
ehuss, line 62: `move` is a keyword, making this invalid syntax.

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -3,7 +3,7 @@
 ## Summary
 
 - `|| a.x + 1` now captures only `a.x` instead of `a`.
-- This can subtly change the drop order of things.
+- This can subtly change the drop order of things, and auto traits usage with closures.
```
rylev, line 6: I'm not sure this is very clear. Maybe the following would be better?
```suggestion
- This can subtly change the drop order of things and whether auto traits are applied to closures or not.
```

src/rust-2021/disjoint-capture-in-closures.md
rylev: Once https://github.com/rust-lang/rust/issues/85894 lands the lint will be renamed, but we can tackle that in a future PR.

src/rust-2021/disjoint-capture-in-closures.md
roxelo: Yes, the expectation is that they would write `let _ = &px;`.

Any thoughts @arora-aman?

arora-aman: We don't usually expect users needing to use `let _ = &px` in a usual case.  

However we should probably add a section saying that this feature/edition 2021 might increase the sizes of their closure and if they are working in a memory critical codebase they can use -Zprofile-closures to get an idea of the effect on size of the closure. In this case, if the closure isn't benefiting from disjoint capture they can force the variables to be completely captured similar to technique used for migration within the closure body.

```rust
let _ = &var;
let _ = (&var1, &var2, &var3);
```

nikomatsakis:
I opened https://github.com/sexxi-goose/edition-guide/pull/2 with some suggested edits

nikomatsakis (approved):

src/rust-2021/disjoint-capture-in-closures.md
```diff
@@ -3,38 +3,154 @@
...
+
+- changes to drop order, or when destructors run ([details](#drop-order));
+- changes to which traits a closure implements ([details](#trait-implementations)).
+
+When running `cargo fix --edition`, Cargo will update the closures in your code to help you migrate to Rust 2021, as described below in [Migrations](#migrations).
+
+Whenever any of the scenarios below are detected, `cargo fix` will insert a "dummy let" into your closure to force it to capture the entire variable:
+
+```rust
+let x = (vec![], vec![]);
```
nikomatsakis, line 43:
```suggestion
let x = (vec![21], vec![22]);
```

nikomatsakis (changes requested):
Make tests pass

nikomatsakis (approved):

PR #248, opened 2021-07-02
Add Rust 2021 prelude migration details

jam1garner:
Currently still needs a bit of work once I have more time ([rendered](https://github.com/jam1garner/edition-guide/blob/master/src/rust-2021/prelude.md))

cc https://github.com/rust-lang/edition-guide/issues/235

### DIFF
--- a/src/rust-2021/prelude.md
+++ b/src/rust-2021/prelude.md
@@ -37,6 +37,12 @@ The tracking issue [can be found here](https://github.com/rust-lang/rust/issues/
 
 As a part of the 2021 edition a migration lint, `rust_2021_prelude_collisions`, has been added in order to aid in automatic migration of Rust 2018 codebases to Rust 2021.
 
+In order to have rustfix migrate your code to be Rust 2021 Edition compatible, run:
+
+```
+cargo fix --edition
+```
+
 ### Implementation Reference
 
 The lint needs to take a couple of factors into account when determining whether or not introducing 2021 Edition to a codebase will cause a name resolution collision (thus breaking the code after changing edition). These factors include:
--- a/src/rust-2021/prelude.md
+++ b/src/rust-2021/prelude.md
@@ -35,7 +35,7 @@ The tracking issue [can be found here](https://github.com/rust-lang/rust/issues/
 
 ## Migration
 
-As a part of the 2021 edition a migration lint, currently `future_prelude_collision`, has been added in order to aid in automatic migration of Rust 2018 codebases to Rust 2021.
+As a part of the 2021 edition a migration lint, `rust_2021_prelude_collisions`, has been added in order to aid in automatic migration of Rust 2018 codebases to Rust 2021.
 
 ### Implementation Reference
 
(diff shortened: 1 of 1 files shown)


src/rust-2021/prelude.md
```diff
@@ -31,5 +31,26 @@ It's identical to the current one, except for three new additions:
...
 
-The library team still needs to formally approve these, which will likely happen soon.
-<!-- TODO: hopefully this happens before we publish this -->
+The tracking issue [can be found here](https://github.com/rust-lang/rust/issues/85684).
+
+## Migration
+
+As a part of the 2021 edition a migration lint, `rust_2021_prelude_collisions`, has been added in order to aid in automatic migration of Rust 2018 codebases to Rust 2021.
+
+### Implementation Reference
```
rylev, line 40: This reference is very interesting, but I'm unsure if this is the level of detail appropriate for this guide. Essentially, the guide should be to help those who are upgrading. This *might* be helpful in the case where someone tries to upgrade and they're confused about why the lint is being triggered, but I don't see that as a major concern. 

I'm personally not against this being here, but if we do keep it, I think I'd like to make it clear to the reader that they most likely don't need this information.

src/rust-2021/prelude.md
```diff
@@ -31,5 +31,26 @@ It's identical to the current one, except for three new additions:
 - [`std::convert::TryFrom`](https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html)
 - [`std::iter::FromIterator`](https://doc.rust-lang.org/stable/std/iter/trait.FromIterator.html)
 
-The library team still needs to formally approve these, which will likely happen soon.
-<!-- TODO: hopefully this happens before we publish this -->
+The tracking issue [can be found here](https://github.com/rust-lang/rust/issues/85684).
+
+## Migration
+
+As a part of the 2021 edition a migration lint, `rust_2021_prelude_collisions`, has been added in order to aid in automatic migration of Rust 2018 codebases to Rust 2021.
```
rylev, line 38: I think most other pages reference `cargo fix --edition` as the mechanism by which the user can perform the automatic migrations. I think we'll want to also say that here.

rylev (changes requested):
I'd like for us to discuss what level of detail is right for the edition guide before merging this.

src/rust-2021/prelude.md
jam1garner: That's fair! Part of my reasoning for making it this detailed (as I wouldn't have otherwise) is that it was expressed by @nikomatsakis that one of the goals should be for users to be able to understand the steps needed to hand-migrate code if the lint fails (assuming I understood correctly)

nikomatsakis: That's right, I think I would probably try to make it a bit more oriented around *solutions* people might try. I can try later to give an example.

nikomatsakis: Opened https://github.com/jam1garner/edition-guide/pull/1 -- what do you think, @jam1garner or @rylev ?

rylev: Add a bunch of comments on that PR - it's certainly the right direction! 😊

nikomatsakis:
Note that @rylev had good suggestions on https://github.com/jam1garner/edition-guide/pull/1 that we still need to apply.

rylev:
I'm going to be taking over this PR, so I've assigned #235 to myself.

rylev:
Closing in favor of #251

PR #249, opened 2021-07-03
reserving-syntax.md: Expand and add detail

joshtriplett:
cc https://github.com/rust-lang/edition-guide/issues/242

### DIFF
--- a/src/rust-2021/reserving-syntax.md
+++ b/src/rust-2021/reserving-syntax.md
@@ -2,44 +2,47 @@
 
 ## Summary
 
-- `any_prefix#..`, `any_prefix".."`, and `any_prefix'..'` are reserved syntax, and no longer tokenize.
+- `any_prefix#`, `any_prefix"..."`, and `any_prefix'...'` are now reserved
+  syntax, and no longer tokenize.
 - This is mostly relevant to macros. E.g. `quote!{ #a#b }` is no longer accepted.
 - It doesn't treat keywords specially, so e.g. `match".." {}` is no longer accepted.
 - Insert whitespace to avoid errors.
 
 ## Details
 
-To make space for some new syntax in the future,
+To make space for new syntax in the future,
 we've decided to reserve syntax for prefixed identifiers and literals:
 `prefix#identifier`, `prefix"string"`, `prefix'c'`, and `prefix#123`,
 where `prefix` can be any identifier.
 (Except those that already have a meaning, such as `b'…'` and `r"…"`.)
+This provides syntax we can expand into in the future without requiring an
+edition boundary. We may use that for temporary syntax until the next edition,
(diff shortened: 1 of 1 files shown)


src/rust-2021/reserving-syntax.md
```diff
@@ -2,44 +2,47 @@
 
 ## Summary
 
-- `any_prefix#..`, `any_prefix".."`, and `any_prefix'..'` are reserved syntax, and no longer tokenize.
+- `any_prefix#`, `any_prefix"..."`, and `any_prefix'...'` are now reserved
```
rylev, line 5: I'm not sure if the use of `any_syntax` is clear. Perhaps we describe it in words (e.g., "the use of any syntax followed by a "#" or literals enclosed in single or double quotes (e.g., `foo#`, `bar'baz'` , `qux"corge"`) is now reserved")

rylev (approved):
Noted a small concern, but nothing blocking merging!

src/rust-2021/reserving-syntax.md
nikomatsakis: I agree that we should at least define `any_prefix`

src/rust-2021/reserving-syntax.md
```diff
@@ -2,44 +2,47 @@
...
+  syntax, and no longer tokenize.
 - This is mostly relevant to macros. E.g. `quote!{ #a#b }` is no longer accepted.
 - It doesn't treat keywords specially, so e.g. `match".." {}` is no longer accepted.
 - Insert whitespace to avoid errors.
```
nikomatsakis, line 9: I think we should expand a bit on migration

nikomatsakis (changes requested):

rylev:
@joshtriplett do you need help getting this merged? We'll soon be doing a sweep over all the sections for consistency, and I'd like this to be merged before we do that.

src/rust-2021/reserving-syntax.md
joshtriplett: I've changed the line in the summary to make it clear that this is an identifier. The subsequent non-summary explanation goes into more detail on that point, so I think this is now addressed.

joshtriplett:
I've updated this to address feedback. I think it's ready to merge now.

rylev (approved):
Looks great! We still need a section on migration, but we can do that in a separate PR.

joshtriplett:
On Thu, Jul 22, 2021 at 01:02:34AM -0700, Ryan Levick wrote:
> Looks great! We still need a section on migration, but we can do that in a separate PR. 

Can you clarify what you're looking for on migration? I tried to cover
what the compiler's migrations would do, and I think other sections of
the edition guide cover "how to invoke migrations".

PR #255, opened 2021-07-15
Update transitioning chapter.

ehuss:
This updates the transitioning chapter to make it more up-to-date, and to correct some issues.

I decided to leave it with a 2018 example, since I think it is clear enough. Most of the 2021 lints are pretty obscure or complicated, and I think readers will get the gist of what it is saying.  I also feel like rewriting this for every edition will be a bit tedious.

This also adds a new chapter to help people if they are having trouble, or need advanced information.

### DIFF
--- a/src/editions/advanced-migrations.md
+++ b/src/editions/advanced-migrations.md
@@ -133,7 +133,7 @@ In this situation, if you end up with code that doesn't work in the next edition
 
 ## Migrating non-Cargo projects
 
-If your project is not using Cargo as a build system, it may still be possible to make use of the automated lints to assist migrating to the next build system.
+If your project is not using Cargo as a build system, it may still be possible to make use of the automated lints to assist migrating to the next edition.
 You can enable the migration lints as described above by enabling the appropriate [lint group].
 For example, you can use the `#![warn(rust_2021_compatibility)]` attribute or the `-Wrust-2021-compatibility` or `--force-warns=rust-2021-compatibility` [CLI flag].
 
--- a/src/editions/advanced-migrations.md
+++ b/src/editions/advanced-migrations.md
@@ -4,8 +4,11 @@
 
 [`cargo fix --edition`][`cargo fix`] works by running the equivalent of [`cargo check`] on your project with special [lints] enabled which will detect code that may not compile in the next edition.
 These lints include instructions on how to modify the code to make it compatible on both the current and the next edition.
-`cargo fix` applies these changes to the source code, and then runs `cargo check` again to verify the fixes work.
+`cargo fix` applies these changes to the source code, and then runs `cargo check` again to verify that the fixes work.
 If the fixes fail, then it will back out the changes and display a warning.
+
+Changing the code to be simultaneously compatible with both the current and next edition makes it easier to incrementally migrate the code.
+If the automated migration does not completely succeed, or requires manual help, you can iterate while staying on the original edition before changing `Cargo.toml` to use the next edition.
 
 The lints that `cargo fix --edition` apply are part of a [lint group].
 For example, when migrating from 2018 to 2021, Cargo uses the `rust-2021-compatibility` group of lints to fix the code.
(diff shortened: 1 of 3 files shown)


src/editions/advanced-migrations.md
```diff
@@ -0,0 +1,173 @@
+# Advanced migration strategies
+
+## How migrations work
+
```
rylev, line 4: Perhaps we can note again that migration works by migrates code to a state that compiles on both the current edition _and_ the next edition. This is why you migrate, then update the edition in Cargo.toml.

rylev (approved):
Looks wonderful!

src/editions/advanced-migrations.md
ehuss: Yea, sounds good. I wanted to make it clear that it is compatible on both editions, but maybe wasn't explicit enough. I added another paragraph trying to explain why it works that way.

ehuss:
I pushed a couple more things that I forgot to include.

src/editions/advanced-migrations.md
```diff
@@ -118,6 +121,28 @@ If your documentation tests use syntax that is not supported in the new edition,
...
+## Generated code
+
+Another area where the automated fixes cannot apply is if you have a build script which generates Rust code at compile time (see [Code generation] for an example).
+In this situation, if you end up with code that doesn't work in the next edition, you will need to manually change the build script to generate code that is compatible.
+
+## Migrating non-Cargo projects
+
+If your project is not using Cargo as a build system, it may still be possible to make use of the automated lints to assist migrating to the next build system.
+You can enable the migration lints as described above by enabling the appropriate [lint group].
+For example, you can use the `#![warn(rust_2021_compatibility)]` attribute or the `-Wrust-2021-compatibility` [CLI flag].
```
rylev, line 133: Shall we suggest using `--force-warn`? (I realize this is not stable yet though)

rylev (approved):
Still looks good. One question though

src/editions/advanced-migrations.md
ehuss: Sure, sounds good!

src/editions/advanced-migrations.md
```diff
@@ -0,0 +1,205 @@
...
+## Migrating macros
+
+Some macros may require manual work to fix them for the next edition.
+For example, a macro that generates syntax that only works on the previous edition, then `cargo fix` will not be able to fix it.
+
+This may be a problem for both [proc macros] and `macro_rules`-style macros.
+`macro_rules` macros can be updated if the macro is used within the same crate, but if it is exported with `#[macro_export]`, then it may not be able to fix it.
+Proc macros in general cannot be fixed at all.
+
+For example, this (contrived) macro won't get fixed when migrating to 2018.
```
nikomatsakis, line 81: I think it'd be good to talk about migrating *crates*, not macros. 

```suggestion
For example, if we migrate the crate containing the (contrived) macro `foo` to 2018, `foo` would not be automatically fixed (note though that macro hygiene, discussed below, means it is possible to migrate other crates that use `foo` to 2018 without changes).
```

src/editions/advanced-migrations.md
```diff
@@ -0,0 +1,205 @@
...
+macro_rules! foo {
+    () => {
+        let dyn = 1;
+        println!("it is {}", dyn);
+    };
+}
+```
+
+If you don't have any tests that actually call this macro, then `cargo fix --edition` won't display any warnings or errors at all.
+However, it won't work when called from another crate.
```
nikomatsakis, line 94:
```suggestion
However, `foo` won't work when called from another crate.
```

src/editions/advanced-migrations.md
nikomatsakis: This text might not actually be right, but I got a bit confused here about the details in reading it, and I imagine others might, too.

nikomatsakis (changes requested):
This looks great. Left one comment about an area I think could be clarified, but I don't consider it a blocker to merging.

src/editions/advanced-migrations.md
ehuss: Thanks, yea after re-reading that section, it was pretty confusing (and misleading in places).  I reworked it to hopefully make it a little clearer and more correct.  

(I noticed that the `keyword_idents` lint doesn't work for macros at all due to the way the parser works, so `cargo fix` won't work even if you call the macro from the same crate.)

nikomatsakis (approved):

PR #267, opened 2021-10-04
Clarify cargo resolver behavior in a workspace.

ehuss:
There was some confusion about how the resolver works in a workspace, so this adds a little clarification.

Closes https://github.com/rust-lang/cargo/issues/9956


rylev (approved):
Looks good. 🎉

PR #273, opened 2021-11-12
Fix a broken link in the Disjoint Capture in Closure chapter

tatsuya6502:
- Fix a broken link to the `UnwindSafe` doc of the standard library reference.
- Remove the second link to the `Clone` doc.


ehuss (approved):
Thanks!

PR #279, opened 2022-08-15
"Creating a new project": add example using 'cargo new --edition YEAR'

salewski:
Also tighten-up existing prose, and use conventional `$` for (non-root) POSIX shell prompt.


ehuss (approved):
Thanks!

PR #282, opened 2023-07-07
b'...' byte strings -> byte chars

w1kee:


ehuss (approved):
Thanks! I'm not sure if the original intended to maybe use double quotes instead, but either choice seems fine.

PR #285, opened 2023-09-06
Update Rust 2018 "Path and module system changes" for Rust 1.72

### LINKED ISSUE #284: Ambiguity rules for `use` have changed in Rust 1.72
The following text at the end of `path-changes`

> If a path is ambiguous, such as if you have an external crate and a local module or item with the same name, you'll get an error, and you'll need to either rename one of the conflicting names or explicitly disambiguate the path.

is no longer true in Rust 1.72 (due to rust-lang/rust#112086).

I'm not sure whether the edition guide is only supposed to care about the behaviour with the latest compiler version.

mattheww:
Closes #284

Since rust-lang/rust#112086, paths in `use` statements have the same shadowing rules as other paths, at least for the simple cases we're talking about in this guide.

The case of an external crate having the same name as a local module still seems worth mentioning, so I've put it in the "Extern crate paths" section.

### DIFF
--- a/src/rust-2018/path-changes.md
+++ b/src/rust-2018/path-changes.md
@@ -272,7 +272,7 @@ enough to have submodules.
 In Rust 2018, paths in `use` declarations and in other code work the same way,
 both in the top-level module and in any submodule. You can use a relative path
 from the current scope, a path starting from an external crate name, or a path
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #292, opened 2024-02-17
Add 2024 prelude

ehuss:


ehuss:
I was thinking about adding a sentence along the lines of "It should be extremely rare that a conflict will happen, because ...", but I wasn't sure exactly how to explain that, and I wasn't certain it is actually true. WDYT?

PR #299, opened 2024-04-11
Flesh out detail behind disallowing refs to static mut

eggyal:
A bit more detail behind the reasoning and alternatives.


ehuss:
Thanks!  @RalfJung or @scottmcm, would you be able to give this a review?

RalfJung:
This makes sense to me.

PR #305, opened 2024-06-08
"cargo fix" now removes "extern crate" statements

tshepang:


ehuss:
I'm not sure this is entirely correct.

Removal of `extern crate` is only done with `--edition-idioms` ([`unused_extern_crates`](https://doc.rust-lang.org/nightly/rustc/lints/listing/allowed-by-default.html#unused-extern-crates) is only part of the idiom group). The lint is also fairly buggy.

I think the correct statement would be to point to the [`unused_extern_crates`](https://doc.rust-lang.org/nightly/rustc/lints/listing/allowed-by-default.html#unused-extern-crates) lint and say it is only available with `cargo fix --edition-idioms` *after* updating the edition to 2018.

traviscross:
Yes, I had this same thought when testing it before merging.

But then I looked at from the perspective of, "is the language being removed wrong enough that it's an improvement to simply remove it?"  The analysis in my head went like this:

- This could not be an edition compatibility lint because it would be an error to remove `extern crate` in Rust 2015.
- So this could only be an edition idiom lint or some other lint that fires on an edition-dependent basis.
- There is now such a lint.
- The edition guide describes edition idiom lints and the process of running them.

So, on that basis, I concluded the language being removed was straightforwardly wrong, as so it was an improvement to simply remove it, and we could follow it up with later PRs to add these kind of caveats where they may be appropriate.

@tshepang: Do you have any interest in proposing a PR with the follow-up that @ehuss mentioned?

PR #310, opened 2024-06-24
Add a section for the never type change in e2024

WaffleLapkin:
I wrote a draft, it should probably be revised.

cc @traviscross


ehuss:
Thanks, can you add it to `SUMMARY.md`? The page won't show up otherwise.

PR #316, opened 2024-08-05
Add chapter for Lifetime Capture Rules 2024

traviscross:
Let's describe the changes to the opaque type lifetime capture rules in some detail.

To do this, we first describe precise capturing, as we'll use that syntax both for describing the meaning of capturing and to describe the migration steps.

Arguably, some of what we describe isn't strictly tied to the edition. People could migrate to the `use<..>` syntax in any edition.  However, this is all part of the same change, both conceptually and practically, so it's better to just describe this in one place.

cc @compiler-errors

Tracking:

- https://github.com/rust-lang/rust/issues/123432

### DIFF
--- a/src/rust-2024/rpit-lifetime-capture.md
+++ b/src/rust-2024/rpit-lifetime-capture.md
@@ -1,12 +1,300 @@
-# RPIT lifetime capture
+# RPIT lifetime capture rules
 
 🚧 The 2024 Edition has not yet been released and hence this section is still "under construction".
 
-This feature is partially implemented, and not yet ready for testing.
-More information may be found in the tracking issue at <https://github.com/rust-lang/rust/issues/117587>.
+This chapter describes changes related to the **Lifetime Capture Rules 2024** introduced in [RFC 3498], including how to use opaque type *precise capturing* (introduced in [RFC 3617]) to migrate your code.
+
+[RFC 3498]: https://github.com/rust-lang/rfcs/pull/3498
+[RFC 3617]: https://github.com/rust-lang/rfcs/pull/3617
 
 ## Summary
 
+- In Rust 2024, *all* in-scope generic parameters, including lifetime parameters, are implicitly captured when the `use<..>` bound is not present.
+- Uses of the `Captures` trick (`Captures<..>` bounds) and of the outlives trick (e.g. `'_` bounds) can be replaced by `use<..>` bounds (in all editions) or removed entirely (in Rust 2024).
+
 ## Details
 
+### Capturing
+
+*Capturing* a generic parameter in an RPIT (return-position impl Trait) opaque type allows for that parameter to be used in the corresponding hidden type.  In Rust 1.82, we added `use<..>` bounds that allow specifying explicitly which generic parameters to capture.  Those will be helpful for migrating your code to Rust 2024, and will be helpful in this chapter for explaining how the edition-specific implicit capturing rules work.  These `use<..>` bounds look like this:
+
+```rust
+# #![feature(precise_capturing)]
+fn capture<'a, T>(x: &'a (), y: T) -> impl Sized + use<'a, T> {
+    //                                ~~~~~~~~~~~~~~~~~~~~~~~
+    //                             This is the RPIT opaque type.
+    //
+    //                                It captures `'a` and `T`.
+    (x, y)
+  //~~~~~~
+  // The hidden type is: `(&'a (), T)`.
+  //
+  // This type can use `'a` and `T` because they were captured.
+}
+```
+
+The generic parameters that are captured affect how the opaque type can be used.  E.g., this is an error because the lifetime is captured despite the fact that the hidden type does not use the lifetime:
+
+```rust,compile_fail
+# #![feature(precise_capturing)]
+fn capture<'a>(_: &'a ()) -> impl Sized + use<'a> {}
+
+fn test<'a>(x: &'a ()) -> impl Sized + 'static {
+    capture(x)
+    //~^ ERROR lifetime may not live long enough
+}
+```
+
+Conversely, this is OK:
+
+```rust
+# #![feature(precise_capturing)]
+fn capture<'a>(_: &'a ()) -> impl Sized + use<> {}
+
+fn test<'a>(x: &'a ()) -> impl Sized + 'static {
+    capture(x) //~ OK
+}
+```
+
+### Edition-specific rules when no `use<..>` bound is present
+
+If the `use<..>` bound is not present, then the compiler uses edition-specific rules to decide which in-scope generic parameters to capture implicitly.
+
+In all editions, all in-scope type and const generic parameters are captured implicitly when the `use<..>` bound is not present.  E.g.:
+
+```rust
+# #![feature(precise_capturing)]
+fn f_implicit<T, const C: usize>() -> impl Sized {}
+//                                    ~~~~~~~~~~
+//                         No `use<..>` bound is present here.
+//
+// In all editions, the above is equivalent to:
+fn f_explicit<T, const C: usize>() -> impl Sized + use<T, C> {}
+```
+
+In Rust 2021 and earlier editions, when the `use<..>` bound is not present, generic lifetime parameters are only captured when they appear syntactically within a bound in RPIT opaque types in the signature of bare functions and associated functions and methods within inherent impls.  However, starting in Rust 2024, these in-scope generic lifetime parameters are unconditionally captured.  E.g.:
+
+```rust
+# #![feature(precise_capturing)]
+fn f_implicit(_: &()) -> impl Sized {}
+// In Rust 2021 and earlier, the above is equivalent to:
+fn f_2021(_: &()) -> impl Sized + use<> {}
+// In Rust 2024 and later, it's equivalent to:
+fn f_2024(_: &()) -> impl Sized + use<'_> {}
+```
+
+This makes the behavior consistent with RPIT opaque types in the signature of associated functions and methods within trait impls, uses of RPIT within trait definitions (RPITIT), and opaque `Future` types created by `async fn`, all of which implicitly capture all in-scope generic lifetime parameters in all editions when the `use<..>` bound is not present.
+
+### Outer generic parameters
+
+Generic parameters from an outer impl are considered to be in scope when deciding what is implicitly captured.  E.g.:
+
+```rust
+# #![feature(precise_capturing)]
+struct S<T, const C: usize>((T, [(); C]));
+impl<T, const C: usize> S<T, C> {
+//   ~~~~~~~~~~~~~~~~~
+// These generic parameters are in scope.
+    fn f_implicit<U>() -> impl Sized {}
+    //            ~       ~~~~~~~~~~
+    //            ^ This generic is in scope too.
+    //                    ^
+    //                    |
+    //     No `use<..>` bound is present here.
+    //
+    // In all editions, it's equivalent to:
+    fn f_explicit<U>() -> impl Sized + use<T, U, C> {}
+}
+```
+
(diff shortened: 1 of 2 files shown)


src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
-This feature is partially implemented, and not yet ready for testing.
-More information may be found in the tracking issue at <https://github.com/rust-lang/rust/issues/117587>.
+This chapter describes changes related to the **Lifetime Capture Rules 2024** introduced in [RFC 3498], including opaque type *precise capturing* introduced in [RFC 3617].
+
+[RFC 3498]: https://github.com/rust-lang/rfcs/pull/3498
+[RFC 3617]: https://github.com/rust-lang/rfcs/pull/3617
 
 ## Summary
 
+- In all editions, `use<..>` bounds allow for capturing only the lifetime parameters needed in RPIT opaque types.
```
compiler-errors, line 12: "needed" is awkward here.

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
+- If a `use<..>` bound is elided (i.e. not present) then the compiler automatically inserts one according to edition-specific rules.
+- In Rust 2024, *all* in-scope generic parameters, including lifetime parameters, are included in these automatically inserted `use<..>` bounds.
```
compiler-errors, line 14: I know that you can get a lot of mileage out of explaining this in terms of `use<..>`, but it also feels obtuse, and possibly less approachable given that people are more familiar with just calling it capturing.

I would probably not frame this in terms of an implicit `use<..>`, because that's literally not what happens in the compiler. Let's just call it capturing?

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
+```rust
+# #![feature(precise_capturing)]
+fn capture<'a, T>(_: &'a (), _: T) -> impl Sized + use<T> {
+    //                                ~~~~~~~~~~~~~~~~~~~
+    //                 The opaque type captures only `T` and not `'a`.
+    ()
+}
+```
+
+You might want to capture more lifetimes in order to return a hidden type that uses those lifetimes.  Conversely, you might want to capture fewer lifetimes in order to allow the opaque type to be used in more places.  E.g., this is an error because the lifetime is captured despite the fact that the hidden type does not use the lifetime:
```
compiler-errors, line 49: > You might want to capture more lifetimes [...]

Would be nice to show an example of how comes about on ed < 2024.

> You might want to capture fewer lifetimes [...]

Would be nice to show an example of how this happens on ed 2024.

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
+```rust
+# #![feature(precise_capturing)]
+fn capture<'a>(_: &'a ()) -> impl Sized + use<> {}
+
+fn test<'a>(x: &'a ()) -> impl Sized + 'static {
+    capture(x) //~ OK
+}
+```
+
+The bounds list of an RPIT opaque type may contain at most one `use<..>` bound.  If this bound is present, then the listed generic parameters are the only ones captured.
```
compiler-errors, line 72: This feels misplaced. I would rather we move this to some reference section on `use`. Or really, we probably don't need to mention this caveat at all?

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
+
+### Edition-specific rules for elided `use<..>` bounds
+
+If the `use<..>` bound is *elided* (i.e., not present), then the compiler automatically inserts one according to edition-specific rules.
```
compiler-errors, line 76: It is unnecessary to frame this in terms of an elided `use<..>` bound. It doesn't bring clarity IMO, over just saying "if `use<..>` is not present, then opaques capture parameters according to edition-specific rules".

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
+
+In all editions, all in-scope type and const generic parameters are added to this automatically-inserted `use<..>` bound.  E.g.:
```
compiler-errors, line 78:
```suggestion
In all editions, all in-scope type and const generic parameters are captured.  E.g.:
```

compiler-errors: e.g.

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
+// In Rust 2024 and later, it's equivalent to:
+//fn f_2024() -> impl for<'a> Tr<'a, Ty = impl Copy + use<'a>> {}
+//                                        ~~~~~~~~~~~~~~~~~~~~
+// However, note that the capturing of higher-ranked lifetimes in
+// nested opaque types is not yet supported.
+```
+
+### Argument position impl Trait (APIT)
+
+Anonymous (i.e. unnamed) generic parameters created by the use of APIT (argument position impl Trait) are considered to be in scope when inserting an explicit `use<..>` bound for an elided one.  E.g.:
```
compiler-errors, line 146:
```suggestion
Anonymous (i.e. unnamed) generic parameters created by the use of APIT (argument position impl Trait) are considered to be in scope for the purposes of capturing.  E.g.:
```

or something like that.

src/rust-2024/rpit-lifetime-capture.md
compiler-errors: I feel like this puts the cart before the horse. I'd probably describe the implcit behavior first, since that's why users are here in the first place, and then describe `use<>` as the convenient fix to all the capturing woes.

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,313 @@
...
+# #![feature(precise_capturing)]
+fn f_elided<T, const C: usize>() -> impl Sized {}
+//                                  ~~~~~~~~~~
+//                       The `use<..>` bound is elided here.
+//
+// In all editions, the above is equivalent to:
+fn f_explicit<T, const C: usize>() -> impl Sized + use<T, C> {}
+```
+
+In Rust 2021 and earlier editions, generic lifetime parameters are *not* included in this automatically-inserted `use<..>` bound for RPIT opaque types in the signature of bare functions and associated functions and methods within inherent impls.  However, starting in Rust 2024, these in-scope generic lifetime parameters *are* included.  E.g.:
```
compiler-errors, line 90: So first, I think we can shorten this by just saying "are *not* always captured". Also, I may have missed it, but do we actually explain *when* lifetimes are captured pre-2024?

compiler-errors (changes requested):

compiler-errors (changes requested):
I still feel like, in general, this doc is doing something weird with the verb "elided". I think most, if not all, usages of the verb "elided" can be replaced with "not present" to the same effect, except it has the added benefit of not acting as if it were something *implicit* to be inserted.

src/rust-2024/rpit-lifetime-capture.md
traviscross: I've remove the word and concept "needed" here, rearranged these bullet points, and removed one of the bullet points about `use<..>`.

src/rust-2024/rpit-lifetime-capture.md
traviscross: I feel like capturing isn't that widely understood, and that people are going to rapidly start thinking of this in a syntactic way in terms of `use<..>`, which is why it was framed this way, but I take the point, and I've reframed it in terms of capturing.

src/rust-2024/rpit-lifetime-capture.md
traviscross: There are examples later in the document when we start talking about the migration.  Maybe we could go on a bit more there, but this chapter is already rather long, and at some point it feels like people should just read RFC 3498 if they want the full motivation.

Capturing is a subtle enough topic that the approach taken here is to introduce and it explain it first before getting into the edition-specific differences in the next section.

src/rust-2024/rpit-lifetime-capture.md
traviscross: I've removed this and the other bits that served to explain `use<..>` syntax.

src/rust-2024/rpit-lifetime-capture.md
traviscross: All uses of the word "elide" or "elided" and this framing have been removed.

src/rust-2024/rpit-lifetime-capture.md
traviscross: All descriptions of automatically adding bounds have been removed.

src/rust-2024/rpit-lifetime-capture.md
traviscross: All descriptions of automatically inserting `use<..>` bounds have been removed.

src/rust-2024/rpit-lifetime-capture.md
traviscross: This paragraph has been changed, as elsewhere, to remove discussion of automatically inserted `use<..>` bounds, and it now states how lifetime parameters are captured in Rust 2021 and earlier editions.

traviscross:
> I still feel like, in general, this doc is doing something weird with the verb "elided"....

The word "elided" has been purged from the document.  Most uses were replaced with "not present" or "no ... is present", and some remaining ones where that wasn't appropriate were replaced with "omit" or "omitted".

src/rust-2024/rpit-lifetime-capture.md
traviscross: The other `use<..>`-related bullet point has now been removed also.

src/rust-2024/rpit-lifetime-capture.md
traviscross: This entire paragraph has now been removed from the document.

src/rust-2024/rpit-lifetime-capture.md
```diff
@@ -1,12 +1,300 @@
...
+# #![feature(precise_capturing)]
+fn f_implicit<T, const C: usize>() -> impl Sized {}
+//                                    ~~~~~~~~~~
+//                         No `use<..>` bound is present here.
+//
+// In all editions, the above is equivalent to:
+fn f_explicit<T, const C: usize>() -> impl Sized + use<T, C> {}
+```
+
+In Rust 2021 and earlier editions, when the `use<..>` bound is not present, generic lifetime parameters are only implicitly captured when they appear syntactically within a bound in RPIT opaque types in the signature of bare functions and associated functions and methods within inherent impls.  However, starting in Rust 2024, these in-scope generic lifetime parameters *are* implicitly captured.  E.g.:
```
compiler-errors, line 75:
```suggestion
In Rust 2021 and earlier editions, when the `use<..>` bound is not present, generic lifetime parameters are only captured when they appear syntactically within a bound in RPIT opaque types in the signature of bare functions and associated functions and methods within inherent impls.  However, starting in Rust 2024, these in-scope generic lifetime parameters *are* unconditionally captured.  E.g.:
```

compiler-errors (approved):
This is kinda awkward still. Otherwise LGTM.

src/rust-2024/rpit-lifetime-capture.md
traviscross: Thanks, updated.

traviscross:
Awkward I can handle.  All 2024 chapters are still a draft, and many need polishing which we'll have some time to do.

PR #320, opened 2024-08-19
2024: Add rustdoc combined doctests

ehuss:
Tracking issue: https://github.com/rust-lang/rust/issues/124853

### DIFF
--- a/src/rust-2024/rustdoc-doctests.md
+++ b/src/rust-2024/rustdoc-doctests.md
@@ -1,0 +1,76 @@
+# Rustdoc combined tests
+
+🚧 The 2024 Edition has not yet been released and hence this section is still "under construction".
+More information may be found in the tracking issue at <https://github.com/rust-lang/rust/issues/124853>.
+
+## Summary
+
+- [Doctests] are now combined into a single binary which should result in a significant performance improvement.
+
+## Details
+
+Prior the the 2024 Edition, rustdoc's "test" mode would compile each code block in your documentation as a separate executable. Although this was relatively simple to implement, it resulted in a significant performance burden when there are a large number of documentation tests. Starting with the 2024 Edition, rustdoc will attempt to combine documentation tests into a single binary, significantly reducing the overhead for compiling doctests.
+
+```rust
+/// Adds two numbers
+///
+/// ```
+/// assert_eq!(add(1, 1), 2);
+/// ```
+pub fn add(left: u64, right: u64) -> u64 {
+    left + right
+}
+
+/// Subtracts two numbers
+///
+/// ```
+/// assert_eq!(subtract(2, 1), 1);
+/// ```
+pub fn subtract(left: u64, right: u64) -> u64 {
+    left - right
+}
+```
+
+In this example, the two doctests will now be compiled in a single executable. Rustdoc will essentially place each example in a separate function within a single binary. Rustdoc also adds a small `main` function which tells it which test to run. The tests still run in independent processes as they did before, so any global state (like global statics) should still continue to work correctly.
+
+This change is only available in the 2024 Edition to avoid potential incompatibilities with existing doctests which may not work in a combined executable.
+
+[doctests]: ../../rustdoc/write-documentation/documentation-tests.html
+[libtest harness]: ../../rustc/tests/index.html
+
+### Standalone tag
+
+In some situations it is not possible for rustdoc to combine examples in a single executable. Rustdoc will attempt to automatically detect if this is not possible, for example:
+
+* Uses the [`compile_fail`][tags] language tag, which indicates that the example fails to compile.
+* Uses the [`edition`][tags], which indicates the edition of the example.[^edition-tag]
+* Uses global attributes like the [`global_allocator`] attribute, which could potentially interfere with other tests.
+* Defines any crate-wide attributes (like `#![deny(unused)]`).
+* Defines a macro that uses `$crate`, because the `$crate` path will not work correctly.
+
+However, rustdoc is not able to automatically determine *all* situations where an example cannot be combined with other examples. In these situations, you can add the `standalone` language tag to indicate that the example should be built as a separate executable.
+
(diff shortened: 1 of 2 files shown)


src/rust-2024/rustdoc-doctests.md
```diff
@@ -0,0 +1,76 @@
...
+///
+/// ```
+/// assert_eq!(subtract(2, 1), 1);
+/// ```
+pub fn subtract(left: u64, right: u64) -> u64 {
+    left - right
+}
+```
+
+In this example, the two doctests will now be compiled in a single executable. Rustdoc will essentially place each example in a separate function within a single binary. Rustdoc also adds a small `main` function which tells it which test to run. The tests still run in independent processes as they did before, so any global state (like global statics) should still continue to work correctly.
```
GuillaumeGomez, line 34: I'm not sure that explaining that rustdoc adds a `main` function is really useful here. Especially since users can add their own `main` function in the doctest directly. I think the sentence can be removed. Just in case, talking about this one:

```
Rustdoc also adds a small `main` function which tells it which test to run.
```

src/rust-2024/rustdoc-doctests.md
```diff
@@ -0,0 +1,76 @@
...
+[libtest harness]: ../../rustc/tests/index.html
+
+### Standalone tag
+
+In some situations it is not possible for rustdoc to combine examples in a single executable. Rustdoc will attempt to automatically detect if this is not possible, for example:
+
+* Uses the [`compile_fail`][tags] language tag, which indicates that the example fails to compile.
+* Uses the [`edition`][tags], which indicates the edition of the example.[^edition-tag]
+* Uses global attributes like the [`global_allocator`] attribute, which could potentially interfere with other tests.
+* Defines any crate-wide attributes (like `#![deny(unused)]`).
```
GuillaumeGomez, line 48: Actually the `deny`, `warn` and `allow` lint levels are not preventing a doctests to be merged. ;)

src/rust-2024/rustdoc-doctests.md
```diff
@@ -0,0 +1,76 @@
...
+* Defines a macro that uses `$crate`, because the `$crate` path will not work correctly.
+
+However, rustdoc is not able to automatically determine *all* situations where an example cannot be combined with other examples. In these situations, you can add the `standalone` language tag to indicate that the example should be built as a separate executable.
+
+```rust
+//! ```
+//! let location = std::panic::Location::caller();
```
GuillaumeGomez, line 55: Perfect example. I think it's the only case where code wasn't working in merged doctest.

src/rust-2024/rustdoc-doctests.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+[tags]: ../../rustdoc/write-documentation/documentation-tests.html#attributes
+[`global_allocator`]: ../../std/alloc/trait.GlobalAlloc.html
+
+[^edition-tag]: Note that rustdoc will only combine tests if the entire crate is Edition 2024 or greater. Using `edition2024` tags in older editions will not result in those tests being combined.
+
+## Migration
+
+There is no automatic migration to determine which doctests need to be annotated with the `standalone` tag. You will need to update your crate to the 2024 Edition and then run your documentation tests and see if any of them fail. If they do fail, you will need to analyze whether the test can be rewritten to be compatible with the combined approach, or add the `standalone` tag to retain the previous behavior.
```
GuillaumeGomez, line 76: I would "tone it down" a bit. In all crates we tested, we only found one case (in `libcore`) which didn't work with merged doctests. I think maybe something "It's very unlikely to happen but some doctests might not worked when merged. The only way to know it is to run them with the 2024 edition. ...". What do you think?

src/rust-2024/rustdoc-doctests.md
```diff
@@ -45,7 +45,7 @@ In some situations it is not possible for rustdoc to combine examples in a singl
 * Uses the [`compile_fail`][tags] language tag, which indicates that the example fails to compile.
 * Uses the [`edition`][tags], which indicates the edition of the example.[^edition-tag]
 * Uses global attributes like the [`global_allocator`] attribute, which could potentially interfere with other tests.
-* Defines any crate-wide attributes (like `#![deny(unused)]`).
+* Defines any crate-wide attributes (like `#![feature(...)]`).
```
GuillaumeGomez, line 48: 👍

src/rust-2024/rustdoc-doctests.md
ehuss: Yea, I was a bit on the fence on how to describe how it works. I was feeling it would be useful to give the reader a better understanding of how things work under the hood. But overall it probably isn't all that important so I have removed it.

src/rust-2024/rustdoc-doctests.md
ehuss: Oh, I didn't catch that, thanks! I switched it to the `feature` attribute which I believe does always prevent it from being combined.

src/rust-2024/rustdoc-doctests.md
GuillaumeGomez: Hum... If it's something people are interested into, would it be ok to link to an external blog post? I explained more in details how it works [here](https://blog.guillaume-gomez.fr/articles/2024-08-17+Doctests+-+How+were+they+improved%3F).

GuillaumeGomez (approved):

GuillaumeGomez:
Looks all good to me, thanks!

src/rust-2024/rustdoc-doctests.md
ehuss: Yea, I thought about adding something along those lines. I added a few sentences to emphasize that it is unlikely to be an issue.

src/rust-2024/rustdoc-doctests.md
ehuss: Sure!

PR #321, opened 2024-09-04
add rustfmt raw identifer sorting doc

calebcartwright:
[rendered](https://github.com/rust-lang/edition-guide/blob/0cbff13a2c44713ca9b7c168d75a225896668cbb/src/rust-2024/rustfmt-raw-identifiers.md)

### DIFF
--- a/src/rust-2024/rustfmt-raw-identifiers.md
+++ b/src/rust-2024/rustfmt-raw-identifiers.md
@@ -1,0 +1,52 @@
+# Rustfmt: raw identifier sorting
+
+🚧 The 2024 Edition has not yet been released and hence this section is still "under construction".
+
+More information may be found in <https://github.com/rust-lang/rust/issues/124764>.
+
+## Summary
+
+`rustfmt` now properly sorts [raw identifiers].
+
+[raw identifiers]: https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html
+
+## Details
+
+The [Rust Style Guide] includes [rules for sorting][sorting] that `rustfmt` applies in various contexts, such as on imports.
+
+Prior to the 2024 Edition, when sorting rustfmt would use the leading `r#` token instead of the ident which led to specious results.
+
+For example:
+
+```rust
+use websocket::client::ClientBuilder;
+use websocket::r#async::futures::Stream;
+use websocket::result::WebSocketError;
+```
+
+Which is now corrected in the 2024 Edition:
+
+```rust
+use websocket::r#async::futures::Stream;
+use websocket::client::ClientBuilder;
+use websocket::result::WebSocketError;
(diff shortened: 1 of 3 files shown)


joshtriplett:
:+1: This looks good.

I don't think this is captured in the style guide, and it should be.

ehuss:
BTW, new pages need to be added to `src/SUMMARY.md`.

I was wondering about the rustfmt changes and whether or not they all need to be on separate pages. I'm unclear about how many changes (from [the list](https://github.com/rust-lang/rust/issues?q=is%3Aopen+is%3Aissue+label%3AC-tracking-issue+label%3AA-edition-2024+rustfmt)) will end up in the edition. If individual changes don't have significant text associated with them, I was wondering if putting them on the same page might make sense (especially if the `Migration` section is identical for each one).

However, if a single page ends up being really long, then that probably wouldn't be good.

Not something we need to decide right now, but would be good to think about.

traviscross:
Yes, I had similar thoughts in both directions, and you described, and ended up suggesting just doing one per page to @calebcartwright.  That makes it a bit easier to follow the flow and template of the other pages.  We can always merge them into one page later if we find it better editorially.

calebcartwright:
Agreed. I'll continue filing these as individual pages but I think _most_ of them would be best merged into one (otherwise I could see a disproportionately high ratio of text in the edition guide doc being "run `cargo fmt`" repeated ad nauseam)

At this point I'm not entirely decided on which ones I think would be best combined on one page, but I'm leaning towards following the model we did for the Style Guide changelog where we had a single entry for "misc rustfmt fixes", which includes this raw ident sorting item.

If it doesn't add any harm/complexity for those of you driving the edition work, perhaps we could just leave these rustfmt PRs open here so that t-style/t-rustfmt can provide input on the text, and then I'll copy pasta things around (with relevant SUMMARY.md entries) and close some of the PRs once we figure out how to structure the full set?

ehuss:
Seems fine to me!

src/rust-2024/rustfmt-raw-identifiers.md
```diff
@@ -0,0 +1,52 @@
...
+Which is now corrected in the 2024 Edition:
+
+```rust
+use websocket::r#async::futures::Stream;
+use websocket::client::ClientBuilder;
+use websocket::result::WebSocketError;
+```
+
+[Rust Style Guide]: https://doc.rust-lang.org/nightly/style-guide/index.html
+[sorting]: https://doc.rust-lang.org/stable/style-guide/index.html?highlight=sort#sorting
```
ehuss, line 36: Just a suggestion to remove the highlighting.

```suggestion
[sorting]: https://doc.rust-lang.org/stable/style-guide/index.html#sorting
```

PR #325, opened 2024-10-01
Rename doctest standalone tag

ehuss:
The `standalone` tag was renamed to `standalone_crate` in https://github.com/rust-lang/rust/pull/130931.


GuillaumeGomez:
Thanks for the update!

PR #326, opened 2024-10-12
2024: Add reserved syntax

ehuss:
This adds a chapter for the reserved syntax in 2024 added in https://github.com/rust-lang/rust/pull/123951 (tracking https://github.com/rust-lang/rust/issues/123735).


traviscross (approved):

traviscross:
@ehuss: I pushed a change to the title and URL to go from "reserving syntax" to "reserved syntax" (see the commit message), and I applied this change to the Rust 2021 guide also and added a redirect.  I'll leave this for you to verify I got the handling of that right.

PR #327, opened 2024-10-14
Update stdout of `cargo new`

muzimuzhi:


src/editions/index.md
```diff
@@ -6,7 +6,7 @@ However, there are times when it's useful to make backwards-incompatible changes
 
 If Rust had suddenly introduced these new keywords, some code would have broken: `let async = 1;` would no longer work.
 
-Rust uses **editions** to solve this problem. When there are backwards-incompatible changes, they are pushed into the next edition. Since editions are opt-in, existing crates won't use the changes unless they explicitly migrate into the new edition. For example, the latest version of Rust doesn't treat `async` as a keyword unless edition 2018 or later is chosen.
+Rust uses **editions** to solve this problem. When there are backwards-incompatible changes, they are pushed into the next edition. Since editions are opt-in, existing crates won't use the changes unless they explicitly migrate into the new edition. For example, the 2015 edition of Rust doesn't treat `async` as a keyword unless edition 2018 or later is chosen.
```
traviscross, line 9: The language here is intentional and shouldn't be changed.  The point is that the latest version of the Rust compiler still has the Rust 2015 behavior unless another edition is explicitly chosen.

traviscross (changes requested):

PR #328, opened 2024-10-23
docs(cargo): Cover MSRV-aware resolver

epage:


ehuss (approved):
Thanks! I applied a small fix for the edition number.

PR #331, opened 2024-11-02
2024: rustfmt style edition

calebcartwright:
https://github.com/rust-lang/rust/issues/123799

[rendered](https://github.com/rust-lang/edition-guide/blob/315514dadac18ef956895359e394d84441df2257/src/rust-2024/rustfmt-style-edition.md)

### DIFF
--- a/src/rust-2024/rustfmt-style-edition.md
+++ b/src/rust-2024/rustfmt-style-edition.md
@@ -1,0 +1,88 @@
+# Rustfmt: Style Edition
+
+🚧 The 2024 Edition has not yet been released and hence this section is still "under construction".
+
+More information may be found in the tracking issue at <https://github.com/rust-lang/rust/issues/123799>.
+
+## Summary
+
+User can now control which style edition to use with `rustfmt`. 
+
+## Details
+
+The default formatting produced by Rustfmt is governed
+by the rules in the [Rust Style Guide].
+
+Additionally, Rustfmt has a formatting stability guarantee that aims
+to avoid causing noisy formatting churn for users when updating a
+Rust toolchain. This stability guarantee essentially means that a newer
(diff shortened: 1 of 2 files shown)


src/rust-2024/rustfmt-style-edition.md
```diff
@@ -0,0 +1,88 @@
...
+Projects who have contributors that may utilize their editor's format-on-save
+features are also strongly encouraged to add a `.rustfmt.toml` file to their project
+that includes the corresponding `style_edition` utilized within their project, or to
+encourage their users to ensure their local editor format-on-save feature is
+configured to use that same `style_edition`.
+
+This is to ensure that the editor format-on-save output is consistent with the
+output when `cargo fmt` is manually executed by the developer, or the project's CI
+process (many editors will run `rustfmt` directly which by default uses the 2015
+edition, whereas `cargo fmt` uses the edition specified in the `Cargo.toml` file)
```
calebcartwright, line 88: I'm curious whether folks think this is relevant enough to include in the guide, and if so, whether this is sufficiently clear?

This is an existing situation for projects that use the format-on-save behavior when working on code that needs edition 2018 or 2021 for rustc parsing, but in those situations the editor would surface up the rustfmt error or do nothing.

With this edition, if a user/project sets the edition to 2024 in their Cargo.toml, and does nothing else, they could trigger a format-on-save from their editor, then manually run a `cargo fmt --check` that would fail because the two were running `rustfmt` with different style editions (or alternatively they run format-on-save in their editor and then push and see their CI fail).

This would only happen if the codebase happens to contain code that gets formatted differently across style editions, but it seems like a sufficiently plausible scenario worth getting in front of in my opinion

ehuss (approved):
Thanks!

PR #332, opened 2024-11-04
2024: rustfmt sorting

calebcartwright:
Related tracking issues:
https://github.com/rust-lang/rust/issues/123800
https://github.com/rust-lang/rust/issues/123802

[rendered](https://github.com/calebcartwright/edition-guide/blob/909ca7c8e464e809e52a8c5301df862cbe6b3cac/src/rust-2024/rustfmt-sorting.md)

A topic we discussed back in #321 was how to bucket the various style/rustfmt related items, many of which are relatively small.

I know we have 3 separate tracking issues covering more detailed levels of the sorting change, but when I put myself in the position of a user consuming this edition guide I'm thinking it would be more intuitive to have a single page/section for the sorting changes


src/rust-2024/rustfmt-sorting.md
```diff
@@ -0,0 +1,43 @@
...
+use std::io::{self, Read, Write, stdout};
+use std::num::{NonZeroU8, NonZeroU16, NonZeroU32, NonZeroU64};
+```
+
+[Rust Style Guide]: ../../style-guide/index.html
+[sorting]: ../../style-guide/index.html#sorting
+
+## Migration
+
+The change can be applied automatically by running `cargo fmt` or `rustfmt` with the 2024 Edition.
```
calebcartwright, line 43: If permissible, I'd like to avoid having to repeat the "here's how to run rustfmt" text in however many style/rustfmt pages we end up having.

Would it suffice to have this one-liner under our migration sections, and then just link out to the detailed instructions for how to do so, which will be added as part of #331?

ehuss:
If you think it is best to have separate pages, that sounds good to me.

Just FYI, eventually we'll reorder and group the pages in SUMMARY.md. Right now they are kinda in random order.

ehuss (approved):
Thanks! I pushed a few fixups. I'm going to go ahead and merge. I'm not sure if there was anything else you wanted to do with this chapter. If so, please follow up with another PR.

PR #337, opened 2024-11-25
Document if/while let chains

est31:
Document if and while let chains, to be stabilized in https://github.com/rust-lang/rust/pull/132833 .

cc tracking issue: https://github.com/rust-lang/rust/issues/53667


Copvt (approved):

ehuss (approved):
Thanks!

PR #345, opened 2024-12-09
Show tail expression temporary example that fails in 2024

### LINKED ISSUE #342: Edition guide for 2024 tail expressions incorrectly imply it only affects drop order within a block, not outside it
Location:
https://github.com/rust-lang/edition-guide/blob/4efbefa611258543294f77ee03643eb16800b399/src/rust-2024/temporary-tail-expr-scope.md

The current content of this page only talks about the drop order *within* a block, and gives an example of code that does not compile in 2021 but does in 2024. However, the 2021 rules allow temporaries to outlive the block into the enclosing expression, such as the `String` in the following program:

```rust
// compiles in 2021; fails in 2024
fn foo() {
    {
        &mut String::new()
    }.push_str("hello");
}
```

This program cannot be fixed merely by reordering or introducing variables within the affected block. This case should be documented in order to help users with situations such as https://github.com/rust-lang/rust/issues/134003. 

@rustbot label Edition-2024

ehuss:
This extends the 2024 tail expression drop chapter to elaborate more on how the temporary scopes have changed. This includes a (contrived) example of something that will fail to compile in 2024, and a suggestion on how to resolve it. I'm not super happy with the example, since it is very unrealistic and unidiomatic, but it is pretty close to lots of real-world examples discovered in crater.

Closes #342


kpreid (approved):
This entirely covers what I reported in #342! Thanks!

PR #351, opened 2024-12-21
2024: Document rustfmt fixes

ehuss:
This is an attempt to document the fixes in the 2024 edition of rustfmt.

I'm not sure this is really the best way to present this. I'm not sure how comprehensible this is.

These are taken from https://github.com/rust-lang/rustfmt/issues/5577. I did not re-review to see if there is anything missing. I probably am missing some things here.

Some of these were a little difficult to explain, since I didn't really understand them. Generally I left the description vague since I didn't think being precise was all that important.

Also, sometimes the code had changed several times since the original introduce of the version two check, and the original examples no longer had the same behavior. All of these examples here were generated with the latest nightly.


traviscross:
cc @rust-lang/rustfmt @rust-lang/style

PR #353, opened 2025-01-07
Rustc{En,De}codable has been removed

RalfJung:
Needed to fix CI in https://github.com/rust-lang/rust/pull/134272 (FCP still pending in that PR)

### DIFF
--- a/src/rust-2024/prelude.md
+++ b/src/rust-2024/prelude.md
@@ -4,7 +4,6 @@
 
 - The [`Future`] and [`IntoFuture`] traits are now part of the prelude.
 - This might make calls to trait methods ambiguous which could make some code fail to compile.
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

TonalidadeHidrica:
Just to make sure, were these description reomved because these macros were entirely removed from Rust in all edition and thus the migration is no longer edition-specific?

RalfJung:
Correct, the macros are entirely gone.

traviscross:
We actually got this a bit wrong.  Since we didn't backport:

- https://github.com/rust-lang/rust/pull/134272

That means that this actually was an edition item, however briefly.  It's probably fine though.

PR #354, opened 2025-01-12
Add alternatives for static-mut-refs

ehuss:
This is an attempt to provide more specific examples on how people can safely work around `static-mut-refs`.

Closes https://github.com/rust-lang/rust/issues/123059


ehuss:
Opening as a draft since this is just a first draft with the intent to solicit feedback and get others involved with figuring out how to guide users on this topic.

traviscross (approved):

PR #358, opened 2025-02-06
Correct spelling error in `static-mut-references.md`

JSorngard:
Thanks for all the work that goes into the guide!
I was reading through the beta version of the guide at https://doc.rust-lang.org/beta/edition-guide/rust-2024/static-mut-references.html#atomics and found this spelling error:

The word "Change" in the first example in the atomics section is spelled "Chnage", this PR fixes that very minor thing.


ehuss (approved):
Thanks!

PR #359, opened 2025-02-17
Fix hyphenation of “Rust-version–aware”

SabrinaJewson:
In standard English grammar, the way to hyphenate “Rust version” and “aware” is by using an en dash (https://en.wikipedia.org/wiki/Dash#Attributive_compounds). Inserting a hyphen between “Rust” and “version” isn’t right since the phrase is not usually hyphenated.

An alternative is to write `rust-version`-aware, i.e. specifically referring to the Cargo key. Not sure if that was the intent here.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -58,7 +58,7 @@
     - [Add `IntoIterator` for `Box<[T]>`](rust-2024/intoiterator-box-slice.md)
     - [Newly unsafe functions](rust-2024/newly-unsafe-functions.md)
   - [Cargo](rust-2024/cargo.md)
-    - [Cargo: Rust version–aware resolver](rust-2024/cargo-resolver.md)
+    - [Cargo: Rust-version–aware resolver](rust-2024/cargo-resolver.md)
     - [Cargo: Table and key name consistency](rust-2024/cargo-table-key-names.md)
     - [Cargo: Reject unused inherited default-features](rust-2024/cargo-inherited-default-features.md)
   - [Rustdoc](rust-2024/rustdoc.md)
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -58,7 +58,7 @@
     - [Add `IntoIterator` for `Box<[T]>`](rust-2024/intoiterator-box-slice.md)
     - [Newly unsafe functions](rust-2024/newly-unsafe-functions.md)
   - [Cargo](rust-2024/cargo.md)
-    - [Cargo: Rust-version aware resolver](rust-2024/cargo-resolver.md)
+    - [Cargo: Rust version–aware resolver](rust-2024/cargo-resolver.md)
     - [Cargo: Table and key name consistency](rust-2024/cargo-table-key-names.md)
     - [Cargo: Reject unused inherited default-features](rust-2024/cargo-inherited-default-features.md)
   - [Rustdoc](rust-2024/rustdoc.md)
--- a/src/rust-2024/cargo-resolver.md
+++ b/src/rust-2024/cargo-resolver.md
@@ -1,8 +1,8 @@
-# Cargo: Rust version–aware resolver
+# Cargo: Rust-version–aware resolver
 
 ## Summary
 
-- `edition = "2024"` implies `resolver = "3"` in `Cargo.toml` which enables a Rust version–aware dependency resolver.
+- `edition = "2024"` implies `resolver = "3"` in `Cargo.toml` which enables a Rust-version–aware dependency resolver.
 
 ## Details
 
@@ -19,7 +19,7 @@ The setting is only honored for the top-level package of the workspace.
 If you are using a [virtual workspace], you will still need to explicitly set the [`resolver` field]
 in the `[workspace]` definition if you want to opt-in to the new resolver.
 
-For more details on how Rust version–aware dependency resolution works, see [the Cargo book](../../cargo/reference/resolver.html#rust-version).
+For more details on how Rust-version–aware dependency resolution works, see [the Cargo book](../../cargo/reference/resolver.html#rust-version).
 
 [`package.rust-version`]: ../../cargo/reference/rust-version.html
 [`resolver.incompatible-rust-version = "fallback"`]: ../../cargo/reference/config.html#resolverincompatible-rust-versions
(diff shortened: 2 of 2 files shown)


ehuss:
@epage Can you review this? I assume the intent was to stay consistent with the key name.

epage:
From my understanding this is changing `Cargo: Rust-version aware resolver` (hyphen) to `Cargo: Rust version–aware resolver` (en-dash) where
- hyphen is intended for joining compound words but is falling out of fashion for space (e.g. fig leaf) or merging the words (e.g. toolbar) ([source](https://en.wikipedia.org/wiki/Hyphen#Use_in_English))
- endash is intended for a compound of a compound ([source](https://en.wikipedia.org/wiki/Dash#Attributive_compounds))

For myself, as a native english speaker, I did not know any of this and I feel this adherence to english rules can get in the way of communication.  I suspect it will be even worse for english-as-a-second-langage.  The visual distinction between a hyphen, an en-dash, an em-dash, etc is very minor and I expect most people will not notice, especially when not next to each other.  I also wonder how many people are like me and are unaware of these rules.  In that context, seeing `Rust version–aware resolver` reads as `rust . version-aware . resolver` vs `rust-version . aware . resolver` or `rust . version . aware . resolver`.  `rust . version-aware . resolver` changes the meaning of the phrase.

We could find an alternative way of phrasing this to avoid this conflict between correctness and clarity but I worry that will lead to a verbosity that will also get in the way.

epage:
Hmm, looking further on [wikipedia](https://en.wikipedia.org/wiki/Hyphen#Compound_modifiers),

> In English, an [en dash](https://en.wikipedia.org/wiki/En_dash), –, sometimes replaces the hyphen in hyphenated compounds if either of its constituent parts is already hyphenated or contains a space (for example, San Francisco–area residents, hormone receptor–positive cells, cell cycle–related factors, and public-school–private-school rivalries).[[29]](https://en.wikipedia.org/wiki/Hyphen#cite_note-LutzStevenson2005-30) A commonly used alternative style is the hyphenated string (hormone-receptor-positive cells, cell-cycle-related factors). (For other aspects of en dash–versus–hyphen use, see [Dash § En dash](https://en.wikipedia.org/wiki/Dash#En_dash).)

It makes it sound like `Rust-version–aware` and `Rust-version-aware` can also be accepted.

SabrinaJewson:
I also think “Rust-version–aware resolver” reads well, so I updated it to that.

salewski:
We seem to be discussing two different things here:

   1. What might be the proper way to hyphenate some combination of words in the sequence "rust version aware resolver"
   2. How to best represent the hyphenation with "characters" (e.g., ASCII/Unicode `HYPHEN-MINUS` (`U+002D`) vs. Unicode `EN DASH` (`U+2013`) vs. something else)

**Regarding (1):** we are describing a "dependency resolver" that is aware of the Rust versions (as opposed to saying the resolver is aware of "versions", abstractly). Therefore the three words "rust version aware" are a single phrasal adjective. Ordinarily, all words in a single phrasal adjective would be hyphenated, so we would have "rust-version-aware resolver".

However, there is a special case for proper nouns, in this case "Rust". Garner[^1] states (under `PHRASAL ADJECTIVES (F)`, p. 830): *'When a name is used attributatively as a phrasal adjective, it ordinarily remains unhyphenated. E.g.: "The Terry Maher strategy..."'.*

As the currently published edition is written, "Rust" is capitalized, meaning that the author was referring  to Rust the programming language, which has assigned versions, rather than `rust-version`, the key name that may appear in a `Cargo.toml` file. If we are to keep the current phrasing, then the proper noun "Rust" should not by connected by a hyphen to the rest of the words in the phrasal adjective, which should themselves be connected. Thus we would have "Rust version-aware resolver".

Alternatively, the clause (currently written as) *"...which enables a Rust-version aware dependency resolver."* could simply be changed to *"...which enables a dependency resolver that is aware of Rust versions."* That would keep "Rust" as a proper noun, but would avoid the phrasal adjective.

There is also a case that could be made for removing the proper noun "Rust". Since the key name in a `Cargo.toml` file is `rust-version`, and that provides the value against which the Rust version comparisons will be made (somebody please correct me if I've got that wrong), it would be reasonable to describe the dependency resolver as being aware of that value. With Markdown, that would be: "`rust-version`-aware resolver".

[^1]: *Garner's Modern English Usage: The Authority on Grammar, Usage, and Style, 5th ed.* by Bryan A. Garner, Oxford University Press, 2022

**Regarding (2):** I don't think Markdown is the proper place to get too deeply into typography, as it is intended to be "used like email" and different renderers will do different things. I would suggest just sticking with ASCII/Unicode `HYPHEN-MINUS` (`U+002D`), which is what you usually get when you type the button next to to zero key on the top row of a typical US keyboard. It is also the expected ASCII character in hyphenated fields in `Cargo.toml` (e.g.: `rust-version`). If we are getting fancier than that, then Markdown is probably not the right tool -- LaTeX would be better for producing pristine typography.

**Summary:**
My recommendations would be:
   1. Got with "`rust-version`-aware resolver"
   2. Stick with ASCII Unicode `HYPHEN-MINUS` (`U+002D`)

Thoughts?

epage:
Note that the Rust version resolver respects the `rust-version` field of your dependency relative to either your `rust-version` field or your Rust toolchain version.   I expect people are considering the last half of this when they see it so being generic about the source of the Rust version would be more correct.  For example, people also refer to this as "MSRV aware resolver", meaning its referring to what your project is doing.

traviscross:
As a resolution here...

Firstly, we're probably not interested in mixing hyphens with unicode en-dashes in this case.

Probably one could argue for "Rust-version-aware resolver".  That would be supportable if it were resolving some ambiguity.  Here, though, I don't think it's really ambiguous.  What we're talking about is a "((Rust version) aware) resolver".  That it, it's a resolver that's aware of the "Rust version", by which we mean the `rust-version` field, the version of the user's Rust toolchain, and the interplay between those.

Given that in "((Rust version) aware) resolver" we're seeking to bind most tightly between the first two tokens, and there's no ambiguity, it's a supportable usage in my view to just minimize the hyphens and say "Rust-version aware resolver", which is what's already there.

So we'll close that.  But thanks for raising this point and for the interesting linguistic discussion.

PR #369, opened 2025-02-24
Fixed a wrong word

iamlockon:
Please take a look at the changes, I believe it's a valid suggestion.

### DIFF
--- a/src/rust-2024/match-ergonomics.md
+++ b/src/rust-2024/match-ergonomics.md
@@ -53,7 +53,7 @@ In Rust 2021 and earlier editions, we allow this oddity:
 let [x, mut y] = &[(), ()]; // x: &(), mut y: ()
 ```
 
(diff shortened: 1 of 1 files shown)


TonalidadeHidrica:
Rather than that, I interpreted it as "pass the reference *to* the pattern" (*from* the scrutinee) because of the word *pass*.

iamlockon:
@TonalidadeHidrica Thanks for the comment. If you read L44 first (naturally), you might feel the same as I do.

traviscross:
Thanks for the suggestion.  What's written is actually what's meant, though this PR has certainly caused us to think about how we could rewrite this sentence overall in a better way.

As a comparison, see the earlier use of the word "pass":

> Here, in the pattern, we pass the outer shared reference in the scrutinee.

So then when we later say:

> Here, because we pass the shared reference in the pattern...

The "in the pattern" is meant in the same way as the "in the pattern" subclause in the last use.

PR #379, opened 2025-11-10
Adjust tail expression temporary scope chapter for new scoping rules

dianne:
This adjusts the wording and examples to accommodate rust-lang/rust#146098.

### DIFF
--- a/src/rust-2024/temporary-tail-expr-scope.md
+++ b/src/rust-2024/temporary-tail-expr-scope.md
@@ -55,35 +55,49 @@ In 2021 the local variable `c` is dropped before the temporary created by `c.bor
 
 ### Temporary scope may be narrowed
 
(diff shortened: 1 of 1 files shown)


src/rust-2024/temporary-tail-expr-scope.md
```diff
@@ -55,35 +55,49 @@ In 2021 the local variable `c` is dropped before the temporary created by `c.bor
 
 ### Temporary scope may be narrowed
 
-When a temporary is created in order to evaluate an expression, the temporary is dropped based on the [temporary scope rules]. Those rules define how long the temporary will be kept alive. Before 2024, temporaries from tail expressions of a block would be extended outside the block to the next temporary scope boundary. In many cases this would be the end of a statement or function body. In 2024, the temporaries of the tail expression may now be dropped immediately at the end of the block (before any local variables in the block).
+When a temporary is created in order to evaluate an expression, the temporary is dropped based on the [temporary scope rules]. Those rules define how long the temporary will be kept alive. Before 2024, temporaries from tail expressions of a block would live past the block to the next temporary scope boundary. In many cases this would be the end of a statement or function body. In 2024, the temporaries of the tail expression may now be dropped immediately at the end of the block (before any local variables in the block).
```
dianne: I'd like to be deliberate about using "extended" only when it applies to temporary lifetime extension. I've reworded instances where it was used to refer to a temporary scope boundary not existing at all.

PR #380, opened 2025-11-15
Link to 1.56.0 and 1.85.0 blog posts

vedranmiletic:
This is consistent with 1.31.0 link in Rust 2018 index page.


ehuss (approved):
Thanks!

PR #383, opened 2026-04-21
ignore tests which will be broken by never type stabilization

WaffleLapkin:
Needed for https://github.com/rust-lang/rust/pull/155499


traviscross (approved):

traviscross:
Thanks @WaffleLapkin.

PR #384, opened 2026-06-12
Update never-type-fallback for never type stabilization

ehuss:
This updates for the never type stabilization proposed in https://github.com/rust-lang/rust/pull/155499.

cc @WaffleLapkin 

This is a draft because the version number will need to be updated.


src/rust-2024/never-type-fallback.md
```diff
@@ -2,17 +2,20 @@
...
-
-[`never_type_fallback_flowing_into_unsafe`]: ../../rustc/lints/listing/deny-by-default.html#never-type-fallback-flowing-into-unsafe
+> [!NOTE]
+> The changes previously documented in this chapter are no longer edition-specific as of Rust CURRENT_RUSTC_VERSION.
+>
+> Never type (`!`) to any type ("never-to-any") coercions now fall back to never type (`!`) rather than to unit type (`()`) on all editions.
+>
+> The `dependency_on_unit_never_type_fallback` lint, which handled the migration, has also been removed.
+>
+> The description of "never-to-any" coercions below is kept for informational purposes.
```
WaffleLapkin, line 12: Would it make sense to leave the rest of the document as-is, instead only mentioning here that it applies only to `Rust < CURRENT_RUSTC_VERSION`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/edition-guide/384/c3c0f0b3da26610138b7ba7663f60cd2c68cf184..f796c483be7c2072a2b905f75443d5ec23950181)*

traviscross (approved):

traviscross:
Thanks @ehuss.  Thanks @WaffleLapkin.

PR #387, opened 2026-08-10
Allow override inherited default-features

0xPoe:
After https://github.com/rust-lang/cargo/pull/17126, we now allow overriding inherited default features. We need to update the document to mention this change.

I’m not sure whether users who migrate from the 2021 edition or earlier to the 2024 edition, especially 1.99+, will be surprised or confused by this behavior change. But as the standard workflow is to run `cargo fix --edition` first, users should notice this change and then decide which behavior they really want. So I hope this change won’t create more chaos here. We just need to clarify the behavior in the corresponding version correctly.

cc: @epage


traviscross (approved):

traviscross:
Thanks @0xPoe.

